PR #8, opened 2020-07-13
feat: update explorer layout to better use space

jake-bassett:


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -21,7 +21,7 @@ import { TableWidgetModel } from './table-widget.model';
     >
       <htc-table
         class="table"
-        [ngClass]="{ 'with-header': this.model.header }"
+        [ngClass]="{ 'header-margin': this.model.header?.topMargin }"
```
anandtiwary, line 24: I have a similar requirement for top-n, hence my ask was to have this supported from titled content. 

Currently, i added margin in the component itself waiting for this to be supported from the model.
https://github.com/hypertrace/hypertrace-ui/blob/c5fa61280229e30ac929cac9032cf258729ffb82/projects/observability/src/shared/components/top-n/top-n-chart.component.scss#L8

Don't want to block this again. So I will approve this PR. We can add the change later to titled content.

jake-bassett: Ah, I see. Yeah, we can change this. Let me get the explore layout stuff through and we can come back and fix. Thanks.

PR #11, opened 2020-07-13
feat: add combo-box and filter tooltips

jake-bassett:


projects/components/src/combo-box/combo-box.component.ts
```diff
@@ -85,6 +85,7 @@ import { ComboBoxMode, ComboBoxOption, ComboBoxResult } from './combo-box-api';
             *ngFor="let option of this.filteredOptions; index as i"
             [class.selected]="this.highlightedOptionIndex === i"
             (click)="this.onOptionClick(option)"
+            [htcTooltip]="option.tooltip ? option.tooltip : ''"
```
aaron-steinfeld, line 88: Is the ternary needed? the tooltip directive should respond to undefined/null/empty string all in the same way

jake-bassett: You are probably right. I didn't even try. I'll change this locally.

PR #16, opened 2020-07-14
refactor: change color palette registration

aaron-steinfeld:


src/app/config.module.ts
```diff
@@ -16,12 +16,11 @@ import { FeatureResolverModule } from './shared/feature-resolver/feature-resolve
...
-      provide: BLUE_COLOR_PALETTE,
-      useValue: ['#001429', '#003149', '#005163', '#007374', '#30947B', '#70B47C', '#B4D17E', '#FFEA8A']
-    },
-    {
-      provide: RED_COLOR_PALETTE,
-      useValue: ['#EEC200', '#F7902D', '#E8654B', '#C44660', '#923768', '#5B2F60', '#27244A', '#001429']
+      provide: DEFAULT_COLOR_PALETTE,
+      useValue: {
+        key: 'default',
+        colors: ['#001429', '#003149', '#005163', '#007374', '#30947B', '#70B47C', '#B4D17E', '#FFEA8A']
```
jake-bassett, line 22: Is there a non hacky way of getting these hex values from css variables? I know we can do it from JS as outlined [here](https://davidwalsh.name/css-variables-javascript) but not sure about doing it from an injector config.

aaron-steinfeld: I'm not aware of any way, and I'd actually argue the palette used for viz should always come from TS and be applied dynamically to allow it to be configured + changed at runtime. In other words, these colors need not live in css at all, at least not in palette form.

PR #17, opened 2020-07-16
feat: update global layout styles

jake-bassett:


projects/components/src/header/page/page-header.component.ts
```diff
@@ -26,10 +27,17 @@ import { IconSize } from '../../icon/icon-size';
           <htc-label [label]="titlecrumb.label"></htc-label>
         </div>
       </div>
+
+      <ng-content></ng-content>
```
aaron-steinfeld, line 31: What's this content expected to be?

projects/components/src/load-async/wrapper/load-async-wrapper.scss
```diff
@@ -1,4 +0,0 @@
-:host {
```
aaron-steinfeld, line 1: this seems like it could cause some issues. Are the loading icons centering in their target space?

projects/components/src/tabs/navigable/navigable-tab.component.ts
```diff
@@ -15,7 +15,17 @@ export class NavigableTabComponent extends ContentHolder {
   @Input()
   public path!: string;
 
-  public get features(): string[] {
+  @Input()
+  public hidden: boolean = false;
+
+  @Input()
+  public features: string[] = [];
```
aaron-steinfeld, line 22: What scenario would these come as inputs?

projects/components/src/tabs/navigable/navigable-tabs.component.ts
```diff
@@ -0,0 +1,18 @@
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { NavigableTab } from './navigable-tab';
+
+@Component({
+  selector: 'htc-navigable-tabs',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <htc-navigable-tab-group>
```
aaron-steinfeld, line 8: This component seems like it does nothing but change the api to to the tab group. Is that something we want to do? We discussed it at the time and designed on the previous API.

projects/distributed-tracing/src/pages/spans/span-list.page.component.ts
```diff
@@ -5,32 +5,32 @@ import { Dashboard, ModelJson } from '@hypertrace/hyperdash';
...
+    >
+    </htc-navigable-dashboard>
   `
 })
 export class SpanListPageComponent implements OnInit {
   public readonly location: string = 'SPANS';
   public readonly scope: string = 'SPAN';
 
+  public readonly filterConfig: NavigableDashboardFilterConfig = {
+    scope: this.scope
```
aaron-steinfeld, line 31: Using the filter through the nav dash is good, but should remove the need for the filter code in this class and the onDashboardReady callback

projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/application-aware-dashboard.component.ts
```diff
@@ -36,6 +35,9 @@ export class ApplicationAwareDashboardComponent implements OnDestroy {
   @Input()
   public editable: boolean = false;
 
+  @Input()
+  public padding?: number;
```
aaron-steinfeld, line 39: Feels real hacky. Can we not pad this through external styles?

projects/distributed-tracing/src/pages/spans/span-list.page.component.ts
jake-bassett: Fixed.

projects/components/src/tabs/navigable/navigable-tabs.component.ts
jake-bassett: I think this API makes sense for how we use navigable tabs now, but I could instead move this code inside page-header.comp.ts and remove this layer.

Will update PR in a moment.

projects/components/src/tabs/navigable/navigable-tab.component.ts
jake-bassett: https://github.com/Traceableai/traceable-ui/pull/1155/files#diff-6ba90f6daf58f0b4c1af9e7debcfb5b9

projects/components/src/header/page/page-header.component.ts
jake-bassett: https://github.com/Traceableai/traceable-ui/pull/1155/files#diff-6ba90f6daf58f0b4c1af9e7debcfb5b9

projects/components/src/load-async/wrapper/load-async-wrapper.scss
jake-bassett: Fixed.

projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/application-aware-dashboard.component.ts
aaron-steinfeld: Agree it's better than overriding the class with ng-deep. I guess this is the tradeoff - bring your own classes and get inconsistency, or build them in and need hacky overrides.

PR #25, opened 2020-07-22
fix: fix truncated text in safari

arjunlalb:

### DIFF
--- a/projects/assets-library/assets/styles/_font.scss
+++ b/projects/assets-library/assets/styles/_font.scss
@@ -92,7 +92,7 @@ $font-family: 'Work Sans', sans-serif;
 @mixin ellipsis-overflow {
   white-space: nowrap;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


projects/assets-library/assets/styles/_font.scss
```diff
@@ -92,6 +92,7 @@ $font-family: 'Work Sans', sans-serif;
 @mixin ellipsis-overflow {
   white-space: nowrap;
   overflow: hidden;
+  overflow: -webkit-paged-x; // For safari
```
aaron-steinfeld, line 95: Can we fix these without browser specifics? Most of these issues are things like safari requires a width when other browsers don't etc.

arjunlalb: will check. I'm also looking at other issues on safari. will see if I can replace this.

PR #27, opened 2020-07-22
fix(color-palette): single color for color palette

skjindal93:
Single color requested does not map to first color in the color palette, since quantize works with count > 1


projects/common/src/color/color-palette.ts
```diff
@@ -16,6 +16,10 @@ export class ColorPalette {
       return [...this.basisColors];
     }
 
+    if (count === 1) {
+      return [this.basisColors[0]];
```
aaron-steinfeld, line 20: What was the previous behavior? Was it giving a mid range value? The modified test indicates it was doing what this is doing.

projects/common/src/color/color-palette.ts
```diff
@@ -16,6 +16,10 @@ export class ColorPalette {
...
+    }
+
     if (count === 2) {
       // Use the middle and end colors for generating 2 colors from a larger palette
       return this.forNColors(3).slice(0, 2);
```
aaron-steinfeld, line 25: If we're reversing the palettes such that we're selecting from the LHS first (which I think makes sense), the 2 count case should swap too, to use start and middle colors.

projects/common/src/color/color-palette.ts
skjindal93: The previous behavior was always outputting the value as `rgb(0, 0, 0)` for `count = 1`, irrespective of the color palette

That's why I changed the test case, since it was using `rgb(0, 0, 0)` as first color in the color palette, which made it pass erroneously

aaron-steinfeld: gotcha, sneaky! also is that GH doing the color rendering? `#FF0000` `rgba(255, 0, 0, 0.5)` `rgb(255,0,0)` Neat, learned something new

projects/common/src/color/color-palette.ts
skjindal93: I think the comment is wrong

It's doing `.slice(0, 2)`, which means it's using the first 2 colors from a larger palette

aaron-steinfeld: It does look that way doesn't it? and the test seems to confirm.
```
  test('should return first and middle color if two colors requested from a larger palette', () => {
    const colors = ['rgb(0, 0, 0)', 'rgb(50, 50, 50)', 'rgb(100, 100, 100)'];
    expect(new ColorPalette(colors).forNColors(2)).toEqual([colors[0], colors[1]]);
  });
```

Could you update that comment then?

skjindal93: yep.

PR #28, opened 2020-07-23
feat: update search box and filter bar styling

jake-bassett:


projects/components/src/combo-box/combo-box.component.ts
```diff
@@ -135,6 +137,12 @@ export class ComboBoxComponent<TValue = string> implements AfterViewInit, OnChan
   @Output()
   public readonly escape: EventEmitter<void> = new EventEmitter();
 
+  @Output()
+  public readonly focused: EventEmitter<void> = new EventEmitter();
```
anandtiwary, line 141: This should emit true/false depending upon what the current state is. It should also emit when the focus and/or blur is cleared out from the component.

projects/components/src/search-box/search-box.component.ts
```diff
@@ -7,18 +7,20 @@ import { IconSize } from '../icon/icon-size';
...
+        class="input"
         type="text"
         [placeholder]="placeholder"
-        class="input"
         [(ngModel)]="value"
-        (input)="onValueChange()"
-        (keyup.enter)="onSubmit()"
+        (input)="this.onValueChange()"
+        (keyup.enter)="this.onSubmit()"
+        (focus)="this.isFocused = true"
```
anandtiwary, line 19: How does `isFocused` gets set to false once the input is in focus?

projects/components/src/combo-box/combo-box.component.ts
jake-bassett: Ah, agreed. that could be a good improvement. Let me change it.

jake-bassett: Hmm. On second thought, while I like that this reduces the API surface area it kind of feels semantically incorrect. Separating focus/blur is very explicit and matches the normal expectations for dom events. If we combine them the API would be suggesting that "not focused" is the same as "blurred" which isn't quite right.

projects/components/src/search-box/search-box.component.ts
jake-bassett: Line 20 should do it.

projects/components/src/combo-box/combo-box.component.ts
aaron-steinfeld: We shouldn't really be mimicing DOM events like this though, should we? Why not use the native events (focusin/out)?

projects/distributed-tracing/src/shared/components/filter-bar/filter-bar.component.ts
```diff
@@ -35,29 +35,29 @@ import { Filter } from './filter/filter-api';
             [scope]="this.scope"
             (apply)="this.onApply($event)"
             (clear)="this.onClear(filter)"
+            (focusin)="this.isFocused = true"
```
aaron-steinfeld, line 38: can we pull these up rather than putting them on each input? Would a focusin on content for example accomplish the same thing with less repetition?

projects/distributed-tracing/src/shared/components/filter-bar/filter-bar.component.ts
```diff
@@ -35,29 +35,29 @@ import { Filter } from './filter/filter-api';
...
+            (focusout)="this.isFocused = false"
           ></htc-filter>
         </div>
 
         <!-- Clear Button -->
-        <htc-button
-          *ngIf="this.internalFilters?.length"
+        <htc-icon
+          class="clear-icon"
```
aaron-steinfeld, line 54: Is this shown even with 0 filters now?

jake-bassett: Good catch. Fix incoming.

projects/components/src/search-box/search-box.component.ts
anandtiwary: oh. I missed that

projects/components/src/combo-box/combo-box.component.ts
anandtiwary: I didn't mean to combine them. I was thinking more like

```
   public readonly focused: EventEmitter<boolean> = new EventEmitter(); 

   public readonly blurred: EventEmitter<boolean> = new EventEmitter(); 
```

PR #29, opened 2020-07-23
test: add improved test tools for dashboards

aaron-steinfeld:


projects/dashboards/src/test/dashboard-verification.ts
```diff
@@ -0,0 +1,32 @@
+import { NavigationService, TimeRangeService } from '@hypertrace/common';
+import { MetadataService } from '@hypertrace/distributed-tracing';
+import { GraphQlRequestService } from '@hypertrace/graphql-client';
+import { ModelJson } from '@hypertrace/hyperdash';
+import { DashboardManagerService } from '@hypertrace/hyperdash-angular';
+import { mockProvider, Spectator } from '@ngneat/spectator/jest';
+import { of } from 'rxjs';
+
+export const isValidDashboard = (
```
anandtiwary, line 9: How is this being used? May be we should rename this to `isValidModelJson` because it seems that's what we are testing?

aaron-steinfeld: It's being used on the page level dashboard tests (didn't update any in this app though). I could go either way on the name - it's not quite either. isValidModeJson doesn't feel quite right because valid json might not produce a valid dashboard. I'll update though.

projects/dashboards/src/test/dashboard-verification.ts
```diff
@@ -0,0 +1,32 @@
...
+  try {
+    spectator.inject(DashboardManagerService).create(json);
+
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+export const mockDashboardProviders = [
```
skjindal93, line 23: So, it's going to be a non-shallow test? Kind of makes sense for dashboards, if we want to check the correctness of the ModelJsons

aaron-steinfeld: Yeah - I'd love for it to just provide the widgets, but the widgets bring along all kinds of services, some of which require mocking due to global config. Providing the widgets ourselves defeats the purpose of the test, so we're stuck with this.

PR #33, opened 2020-07-24
feat: support table page in url

aaron-steinfeld:


projects/components/src/table/table.component.ts
```diff
@@ -422,4 +443,31 @@ export class TableComponent
...
+        pageSize: this.paginator.pageSize,
+        pageIndex: this.paginator.pageIndex
+      });
+    }
+  }
+
+  private pageDataFromUrl(params: ParamMap): Partial<PageEvent> | undefined {
+    return this.syncWithUrl
+      ? {
+          pageSize: new NumberCoercer().coerce(params.get(TableComponent.PAGE_SIZE_URL_PARAM)),
```
anandtiwary, line 468: nit: we could use the same coercer object.

projects/components/src/table/table.component.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    component: TableComponent,
+    shallow: true,
+    imports: [LetAsyncModule],
+    providers: [
+      mockProvider(NavigationService),
+      mockProvider(ActivatedRoute),
+      mockProvider(ActivatedRoute, {
+        queryParamMap: EMPTY
+      }),
+      mockProvider(NavigationService)
```
anandtiwary, line 35: Should we also mock tooltip/popover here? popover depends on `navigation$` which would be undefined here, hence throwing that console error.

projects/components/src/table/table.component.ts
aaron-steinfeld: didn't seem worth it to me, felt It read cleaner this way rather than storing it in a var.

projects/components/src/table/table.component.test.ts
aaron-steinfeld: popover isn't part of this test, so not required.Beauty of shallow tests.

anandtiwary: But tooltip is. The console warnings I saw were due to the display of tooltips which in turn depends on popover. If it is being initialized in test then it would likely throw the warning (either now or with a future change). I was thinking we should mock the directive to prevent this?

aaron-steinfeld: Tooltip's not part of this test either - it's shallow, only things explicitly declared here are part of the test.

anandtiwary: One of the test I was playing with was still throwing the console warning after I converted it into shallow test. Not sure why, may be I did something wrong. Still learning about shallow tests :)

aaron-steinfeld: So shallow only says - don't throw if a directive/component is missing. If you mark a test as shallow but still import tooltip, then it's not actually shallow. The situation you describe sounds like it was still importing tooltip - here the only real directive we're using is letasync (since structural directives are a lot more painful to mock)

projects/components/src/paginator/paginator.component.ts
```diff
@@ -64,12 +64,24 @@ export class PaginatorComponent implements OnChanges, PaginationProvider {
...
+    this._pageIndex = pageIndex;
+  }
+  public get pageIndex(): number {
+    return this._pageIndex ?? 0;
+  }
+  private _pageIndex?: number;
 
   @Input()
-  public pageSize: number = 50;
+  public set pageSize(pageSize: number) {
```
anandtiwary, line 76: If I pass a random page size, would it appear on the available page size options dropdown?

projects/components/src/table/table.component.test.ts
anandtiwary: I see.

projects/components/src/paginator/paginator.component.ts
aaron-steinfeld: It probably would not - although that's an existing issue in component, it's a little more obvious now. I can probably merge in the actual pageSize with the options if we want to - or reject a page size that's not in the options.

anandtiwary: Yeah. I was thinking we should merge it in the available options. Rejecting it would make it a little rigid.

aaron-steinfeld: OK, will add that in - mind if it's a separate PR?

anandtiwary: No Sir!

PR #42, opened 2020-07-27
feat: add table sort in url and enable for full page tables

aaron-steinfeld:


projects/components/src/table/table.component.ts
```diff
@@ -149,13 +151,16 @@ export class TableComponent
...
     AfterViewInit,
+    OnDestroy,
     ColumnConfigProvider,
     TableDataSourceProvider,
     FilterProvider,
     ColumnStateChangeProvider,
     RowStateChangeProvider {
   private static readonly PAGE_INDEX_URL_PARAM: string = 'page';
   private static readonly PAGE_SIZE_URL_PARAM: string = 'page-size';
+  private static readonly SORT_COLUMN_URL_PARAM: string = 'sort';
```
anandtiwary, line 162: sort-attribute? or sort-on ?

projects/components/src/table/table.component.ts
```diff
@@ -453,12 +471,15 @@ export class TableComponent
     }
   }
 
-  private initializeUrl(): void {
-    if (this.paginator) {
-      this.onPageChange({
```
anandtiwary, line 458: How is this being done now?

projects/components/src/table/table.component.ts
aaron-steinfeld: updated to `sort-by` based on offline conversation

projects/components/src/table/table.component.ts
aaron-steinfeld: It didn't feel valuable to put the default page info in the url on load. I can buy that argument for something like the selected tab, but for page size I think the simpler url for the more common use case is better.

anandtiwary: Okay I agree with that.

PR #43, opened 2020-07-29
feat: filter tabs

jake-bassett:

### DIFF
--- a/projects/assets-library/assets/styles/_layout.scss
+++ b/projects/assets-library/assets/styles/_layout.scss
@@ -47,28 +47,33 @@
   }
 }
 
-@mixin card-hover {
-  &:hover {
-    @include box-shadow();
-  }
-}
-
-@mixin box-shadow {
-  box-shadow: 0 1px 4px rgba($gray-7, 0.16), 0 3px 4px rgba($gray-7, 0.08), 0 2px 5px rgba($gray-7, 0.16);
-}
-
 @mixin frame-border {
   border: 1px solid $color-border;
 }
 
+@mixin card-hover {
+  &:hover {
+    @include dropdown-box-shadow();
+  }
+}
+
+@mixin filter-box-shadow {
+  box-shadow: 2px 2px 2px rgba(225, 228, 229, 0.5), 2px -2px 4px rgba(225, 228, 229, 0.2),
+    0px 2px 4px rgba(225, 228, 229, 0.2);
+}
+
+@mixin dropdown-box-shadow {
+  box-shadow: 0 1px 4px rgba($gray-7, 0.16), 0 3px 4px rgba($gray-7, 0.08), 0 2px 5px rgba($gray-7, 0.16);
+}
+
 @mixin dropdown($radius: 4px) {
-  @include box-shadow;
+  @include dropdown-box-shadow();
   width: inherit;
   border-radius: $radius;
   background: white;
   overflow-y: auto;
   overflow-x: hidden;
   position: relative;
-  left: 1px; // To visually center accounting for margin
+  left: 1px; // To visually center while accounting for margin
   margin: 1px 0; // Keeps the dropdown from covering the trigger border
 }
--- a/projects/components/src/filter-tabs/filter-tabs.component.scss
+++ b/projects/components/src/filter-tabs/filter-tabs.component.scss
@@ -1,0 +1,46 @@
+@import 'layout';
+@import 'font';
+@import 'color-palette';
+
+.filter-tabs {
+  @include flex-layout(row);
+  @include body-1-medium($gray-5);
+  position: relative;
+  border: 1px solid $color-border;
+  border-radius: 6px;
+  background: $gray-1;
+  height: 36px;
+
+  .container {
+    display: flex;
+  }
+
+  .divider {
+    width: 1px;
+    margin: 6px 1px;
+    background: $gray-2;
+
+    &.adjacent-active {
+      background: $gray-1;
+    }
+  }
+
+  .active {
+    @include center-contents();
+    @include filter-box-shadow();
+    height: calc(100% - 4px); // Subtract 4px for margin
+    margin: 2px 0;
+    position: absolute;
+    width: 0; // Modified in html/ts
+    left: 0; // Modified in html/ts
+    background: white;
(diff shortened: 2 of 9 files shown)


projects/assets-library/assets/styles/_layout.scss
```diff
@@ -47,28 +47,33 @@
...
 }
 
+@mixin frame-border {
+  border: 1px solid $color-border;
+}
+
 @mixin card-hover {
   &:hover {
-    @include box-shadow();
+    @include dropdown-box-shadow();
```
anandtiwary, line 56: Can we please move these to its own interaction.scss?

projects/components/src/filter-tabs/filter-tabs.component.ts
```diff
@@ -0,0 +1,94 @@
...
+        [style.width.px]="this.activeElementPosition.width"
+      ></div>
+      <div class="container" *ngFor="let tab of this.tabs; let index = index">
+        <div class="divider" *ngIf="index !== 0" [class.adjacent-active]="this.isAdjacentActiveTab(index)"></div>
+        <htc-filter-tab class="tab" [label]="tab.label" (click)="this.setActiveTab(tab)"></htc-filter-tab>
+      </div>
+    </div>
+  `
+})
+export class FilterTabsComponent implements AfterViewInit, OnChanges {
```
anandtiwary, line 37: Is this going to replace toggle filter?

projects/components/src/filter-tabs/filter-tabs.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  public ngAfterViewInit(): void {
+    this.setActiveTab();
+  }
+
+  public setActiveTab(tab?: FilterTab): void {
+    this.activeTab = tab || (this.tabs && this.tabs[0]);
+    this.activeTabIndex = this.tabs?.indexOf(this.activeTab!);
+    this.activeElementPosition = this.getElementPosition(this.activeTab!, this.activeTabIndex);
+    this.activeTabChange.emit(this.activeTab);
+    this.changeDetector.markForCheck();
```
anandtiwary, line 70: Wouldn't this through View changed after check error?

anandtiwary: We could probably use QueryListAndChanges

projects/components/src/filter-tabs/filter-tabs.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  public isAdjacentActiveTab(index: number): boolean {
+    return (
+      this.activeTabIndex !== undefined && (index - this.activeTabIndex === 1 || index - this.activeTabIndex === 0)
```
anandtiwary, line 75: the rhs of second condition should also be compared against 1 isn't?

projects/components/src/filter-tabs/filter-tabs.component.ts
aaron-steinfeld: Or just make the active tab an observable

projects/components/src/filter-tabs/filter-tabs.component.scss
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  .active {
+    @include center-contents();
+    @include filter-box-shadow();
+    height: calc(100% - 4px); // Subtract 4px for margin
+    margin: 2px 0;
+    position: absolute;
+    width: 0; // Modified in html/ts
```
aaron-steinfeld, line 34: This is a simple tab, it does not seem from an outside look like a situation where we need to use absolute, measure, set explicit widths etc. What's preventing this from being done using more simple layout?

projects/assets-library/assets/styles/_layout.scss
jake-bassett: https://traceableai.atlassian.net/browse/ENG-5455

projects/components/src/filter-tabs/filter-tabs.component.ts
jake-bassett: Toggle Button? No, this component is specifically for toggle filters. The toggle buttons will still exist but won't be used for filters anymore.

projects/components/src/filter-tabs/filter-tabs.component.ts
jake-bassett: Not sure what you mean. What we want to do here is remove the divider lines surrounding the active filter so we compare our index to the active one. If the diff is 0 or 1 we know we are next to the active tab.

projects/components/src/filter-tabs/filter-tabs.component.scss
jake-bassett: This is all to support the animations. The animation requires the movement of an element. But we only want the white background to move back and forth (left position) and adjust its width to match the parent tab.

jake-bassett:
> My high level feedback - what is this component's purpose compared to the toggle button? This doesn't seem at all related to filters, even if that's one use case for it. I don't think we need both things, so maybe whatever functionality or styling that is missing from toggle can be added?

This component is _only_ for filters. The toggle buttons are still a thing, but they will not be used for filters anymore.

projects/components/src/filter-tabs/filter-tabs.component.ts
```diff
@@ -0,0 +1,90 @@
...
+      </div>
+    </div>
+  `
+})
+export class FilterTabsComponent implements AfterViewInit, OnChanges {
+  @Input()
+  public readonly tabs?: FilterTab[] = [];
+
+  @Input()
+  public activeTab?: FilterTab;
```
jake-bassett, line 41: This should be readonly. Fix incoming.

projects/components/src/filter-tabs/filter-tabs.component.ts
```diff
@@ -0,0 +1,90 @@
...
+      this.setActiveTab();
+    }
+  }
+
+  public ngAfterViewInit(): void {
+    this.setActiveTab();
+  }
+
+  public setActiveTab(tab?: FilterTab): void {
+    this.activeTab = tab || (this.tabs && this.tabs[0]);
```
jake-bassett, line 63: Should not reassign an Input. Fix incoming.

PR #46, opened 2020-07-30
Toggle group

jake-bassett:


projects/components/src/toggle-group/toggle-group.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    const activeItem: ToggleItem = item || this.items[0];
+
+    this.activeItemIndex = this.items.indexOf(activeItem);
+    this.activeElementPosition = this.getElementPosition(activeItem, this.activeItemIndex);
+    this.activeItemChange.emit(activeItem);
+  }
+
+  public isAdjacentActiveItem(index: number): boolean {
+    return (
+      this.activeItemIndex !== undefined && (index - this.activeItemIndex === 1 || index - this.activeItemIndex === 0)
```
aaron-steinfeld, line 86: This was asked in the other PR but I didn't understand the explanation - an adjacent index would be either one greater or one less than the active item, but we seem to be checking if it's one greater or equal.

projects/components/src/toggle-group/toggle-group.component.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ToggleGroupComponent, ToggleGroupModule, ToggleItem } from '@hypertrace/components';
+import { byText, createHostFactory, mockProvider, Spectator } from '@ngneat/spectator/jest';
+import { EMPTY } from 'rxjs';
+
+describe('Toggle Group Component', () => {
+  let spectator: Spectator<ToggleGroupComponent>;
+
+  const createHost = createHostFactory({
+    component: ToggleGroupComponent,
+    declareComponent: false,
```
aaron-steinfeld, line 12: Please convert this to a shallow test and remove the import - that should also remove the need to mock nav service, which I don't see being used by this test.

projects/components/src/toggle-group/toggle-group.component.test.ts
```diff
@@ -0,0 +1,57 @@
...
+          activeItemChange: activeItemChangeSpy
+        }
+      }
+    );
+
+    spectator.tick();
+    expect(activeItemChangeSpy).toHaveBeenCalledWith(items[1]);
+
+    spectator.click(spectator.query(byText('First'))!);
+    spectator.tick();
```
aaron-steinfeld, line 54: Is this tick needed? The only thing async I see in here is the after view init accounted for on line 50.

projects/components/src/toggle-group/toggle-group.component.ts
jake-bassett: Maybe the name I chose isn't the best, as usual. :-)

Each item owns its left side vertical divider bar. This method is called by each item to see if it needs to remove the divider it owns. From an item perspective, if we are active or the item to our left is active, we want to remove our divider.

projects/components/src/toggle-group/toggle-group.component.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import { byText, createHostFactory, Spectator } from '@ngneat/spectator/jest';
+import { ToggleItemComponent } from './toggle-item.component';
+
+describe('Toggle Group Component', () => {
+  let spectator: Spectator<ToggleGroupComponent>;
+
+  const createHost = createHostFactory({
+    component: ToggleGroupComponent,
+    shallow: true,
+    declarations: [ToggleItemComponent, LabelComponent]
```
aaron-steinfeld, line 12: Why does this need the real label component? I'm guessing it's because we're verifying the text through how the label component renders it. Instead, in a shallow test, bests practice would be to mock it, and verify the label's input is set to whaat you expect. This means that if label were to change (say, add a new component internally) this test won't break. See the paginator test if you'd like an example - that mocks label.

PR #48, opened 2020-07-31
Interaction mixins

jake-bassett:


projects/components/src/combo-box/combo-box.component.scss
```diff
@@ -1,6 +1,4 @@
-@import 'font';
-@import 'color-palette';
-@import 'layout';
+@import 'mixins';
```
anandtiwary, line 1: Why this? Shouldn't we only import those mixins which are actually used inside the scss? There might be some additional cost to it. I remember reading something about it in the article which SJ shared.

jake-bassett: Mainly convenience, but if there is a performance cost then I'm cool just importing individual files. 

I thought the whole reason we discussed and were fine with importing mixins everywhere in the first place was because they are tree shaken out.

aaron-steinfeld: Runtime it should be the same, but we can verify. Not sure if it has any significant compile time issues. My personal preference is the more explicit imports as it gives more of a sense of dependencies (else, I wouldn't even bother separating them in different files), but this isn't something I feel super strong on, so will defer (that's my new thing!)

jake-bassett: If there is no downside, I pretty strongly prefer the single import. It is way easier to only have to know to import the one file and its also less maintenance as we continue refactor the global mixin files. 

That said, if you guys both strongly prefer explicit, I can change back.

aaron-steinfeld: That's fine - I'll wait for anand to chime in, but I'm good with it.

anandtiwary: I can't do a poc on perf now :) . Making it explicit makes it easier to read and we are only importing what we need. But I see Jake's point regarding the advantages. Don't have any strong feeling so we can get this in.

PR #56, opened 2020-08-03
feat(table): multi row select

skjindal93:


anandtiwary:
Are we also planning to show a header checkbox for selecting/deselecting all rows in the page?

skjindal93:
> Are we also planning to show a header checkbox for selecting/deselecting all rows in the page?

May be in future. The current use case definitely does not need that

projects/components/src/table/data/table-cdk-column-util.ts
```diff
@@ -8,7 +8,12 @@ export namespace TableCdkColumnUtil {
   ): changedColumn is TableColumnConfig => changedColumn !== undefined;
 
   export const removeNonDataConfigs = (columnConfigs: TableColumnConfig[]): TableColumnConfig[] =>
-    columnConfigs.filter(columnConfig => columnConfig.renderer !== StandardTableCellRendererType.RowExpander);
+    columnConfigs.filter(
+      columnConfig =>
+        columnConfig.renderer !== StandardTableCellRendererType.RowExpander &&
+        columnConfig.renderer !== StandardTableCellRendererType.Checkbox &&
+        columnConfig.visible
```
aaron-steinfeld, line 15: Why only filter visible columns?

projects/components/src/table/table.component.ts
```diff
@@ -390,8 +418,9 @@ export class TableComponent
   }
 
   public toggleRowSelection(row: StatefulTableRow): void {
-    this.selection = this.selection === row ? undefined : row;
-    this.selectionChange.emit(this.selection);
+    row.$$state.selected = !row.$$state.selected;
+    this.selections = find(this.selections, row) ? remove(this.selections, row) : this.selections.concat(row);
```
aaron-steinfeld, line 422: can we just use contains (or equivalent) since we're not using the find result other than truthiness

projects/components/src/table/table.component.ts
```diff
@@ -169,6 +170,21 @@ export class TableComponent
     onClick: (row: StatefulTableRow) => this.toggleRowExpanded(row)
   };
 
+  private readonly selectableRowColumnConfig: TableColumnConfig = {
+    field: '$$state',
```
aaron-steinfeld, line 174: - Why is this single selectable column needed?
- Is it problematic that we're using $$state for multiple columns? I thought we need field to be unique.
- The naming should more clearly delineate that selectablecolumn and checkbox column are both selection modes.

jake-bassett: I _believe_ this will work because we only have one $$state column at a time... i.e. we can't have expandable rows with a checkbox, but that should be fine.

projects/components/src/table/table.component.ts
```diff
@@ -204,7 +220,7 @@ export class TableComponent
   public initialExpandAll?: boolean = false;
 
   @Input()
-  public selection?: StatefulTableRow;
+  public selections: StatefulTableRow[] = [];
```
jake-bassett, line 223: We can't always guarantee the user won't pass undefined here.

projects/components/src/table/table.component.ts
aaron-steinfeld: but we can have expandable rows with single select right? Both of those use $$state

jake-bassett: Hmm, which screen is that?

Anyway, I actually think this will work now that I'm remembering some more. Don't think the `field` needs to be unique after all. If I remember correctly, its just used to pluck the relevant data from the row dictionary and hand that to the cell renderer. In which case, this will work for both selection and expansion.

But I'll let @skjindal93 confirm. I'm sure he tested. :-)

skjindal93: It needs to be unique. Otherwise, it throws "Duplicate column definition name provided"

I made expandable and multi select mutually exclusive. Because, there aren't use cases yet.

jake-bassett: @aaron-steinfeld did you see a use case for this in the current mocks?

skjindal93: > Why is this single selectable column needed?

Since, I added a new property `selected` in `TableRowState` and I wanted to maintain the `selected` property even for single selection mode (_to make it look complete_)

Hence, added an invisible single selectable column to maintain state

projects/components/src/table/data/table-cdk-column-util.ts
skjindal93: because, I assumed that visible columns would be the data fields, otherwise, why would they be invisible?

And, I needed this for adding a single selectable column state, but excluding it from the data column configs

aaron-steinfeld: OK I looked where it was used, makes sense. Maybe worth a rename though (since these are still data configs)

projects/components/src/table/table.component.ts
aaron-steinfeld: > Hmm, which screen is that?

Any expandable screen really, right? Since single select is the default. The mutual exclusivity is only between the multiselect mode.


> Since, I added a new property selected in TableRowState and I wanted to maintain the selected property even for single selection mode (to make it look complete)


It makes sense to maintain it, but I'm not following how the column helps that

skjindal93: > It makes sense to maintain it, but I'm not following how the column helps that

I thought there was a 1-1 mapping between the field `'$$state'` and the property defined in the StatefulTableRow interface

```
export interface StatefulTableRow extends TableRow {
  $$state: TableRowState;
}
```

Turns out, I was wrong

skjindal93: > Any expandable screen really, right? Since single select is the default. The mutual exclusivity is only between the multiselect mode.

The single select still happens on the expandable screen by clicking on rows (_if that's what you are referring to_)

projects/components/src/table/data/table-cdk-column-util.ts
skjindal93: rename of?

projects/components/src/table/table.component.ts
aaron-steinfeld: > The single select still happens on the expandable screen by clicking on rows (if that's what you are referring to)
Right, but those screens would have two $$state columns - anyway, given the comment above about it not needing to be 1-1, it sounds like we can remove it and this conversation is moot.

projects/components/src/table/data/table-cdk-column-util.ts
aaron-steinfeld: this function - a hidden column config is still a data config. Instead of `removeNonDataConfigs`, it could instead be `fetchableColumnConfigs` or the like.

PR #57, opened 2020-08-03
style: update toggle transition

jake-bassett:


projects/components/src/toggle-group/toggle-group.component.scss
```diff
@@ -1,8 +1,12 @@
 @import 'mixins';
 
+:host {
+  display: flex;
```
aaron-steinfeld, line 4: Why is this needed? It feels like this is our default escape hatch whenever layout isn't working, but it just leads to us using it more and more.

jake-bassett: I'm still not sure why we are against `:host` so much. I can fully understand why we have to justify `ng-deep` every time we use it, but I don't know of any ill effects for using `:host`. Its just putting a class on the wrapping element that Angular creates.

Anyway, adding a flex layout to host allows the component to size to its contents instead of expanding to fill whichever ancestor container holds the layout stack. Without a parent flex container somewhere, the whole grouping expands until some ancestor stops it.

I didn't try, but I could also probably do this by adding another `div` to surround the current template if we have a good reason not to use `:host`.

aaron-steinfeld: > I didn't try, but I could also probably do this by adding another div to surround the current template if we have a good reason not to use :host.

This is generally what I do when we have this requirement, but it sounds like we're not on the same page WRT `:host` so let's discuss that first.

My issue with `:host` is that it changes expectations on the consumer - in both directions.

So by default, the expectation is that an element renders as inline - :host changes that expectation, and makes you have to look up the implementation for each element you use. This problem cuts in both directions. If I rely on the styles on a host element for my own layout, they can later be changed by the component author. Relying on that layout is not only more fragile, but also restrictive. For example, we often set width to 100% on host, which means I can't set a fixed width externally. In the other direction, styles I set on the host could break the implementation of the component (in this case, if I set display to inline, any flex children would render differently).

So long story short, the host in my mind serves as a DMZ so to speak, creating a decoupling between the component's implementation and usage. It makes a component more robust to change without affecting all usages of that component.

PR #58, opened 2020-08-04
feat: update type for ToggleItem value

jake-bassett:


projects/components/src/toggle-group/toggle-item.ts
```diff
@@ -1,4 +1,4 @@
-export interface ToggleItem<T = unknown> {
+export interface ToggleItem<T = unknown | undefined> {
```
aaron-steinfeld, line 1: undefined is part of unknown, right?

PR #59, opened 2020-08-04
fix: update timestamp display to use uniform number of chars

aaron-steinfeld:


projects/common/src/utilities/formatters/date/date-formatter.ts
```diff
@@ -54,7 +54,7 @@ export class DateFormatter {
         return 'd MMM y h:mm a';
       case DateFormatMode.DateAndTimeWithSeconds:
       default:
-        return 'y-M-d hh:mm:ss a';
+        return 'y-MM-dd hh:mm:ss a';
```
jake-bassett, line 57: Oh thank god.

Is the cell renderer using this so it will be fixed as well?

aaron-steinfeld: sure is

aaron-steinfeld: sure is.

anandtiwary: sure is..

PR #60, opened 2020-08-04
Revert "feat: update type for ToggleItem value"

jake-bassett:
Reverts hypertrace/hypertrace-core-ui#58


projects/components/src/toggle-group/toggle-item.ts
```diff
@@ -1,4 +1,4 @@
 export interface ToggleItem<T = unknown> {
   label: string;
-  value: T;
+  value?: T;
```
anandtiwary, line 3: nit: The PR summary and change doesn't match.

PR #61, opened 2020-08-05
feat(table): remove selections from rows on change

skjindal93:


projects/components/src/table/table.component.ts
```diff
@@ -278,6 +279,15 @@ export class TableComponent
       this.columnConfigsSubject.next(this.buildColumnConfigs());
     }
 
+    if (changes.selections && this.selections) {
+      // unmark the selections which don't exist in existing row selections
+      this.rowSelections.forEach(row => {
+        this.selections!.includes(row) ? row.$$state.selected = true : row.$$state.selected = false
```
aaron-steinfeld, line 285: Please flip the ternary - no expressions as assignments. But aside from that, don't we need to set $$state.selected for all rows? this would unset it in an existing selection, but would it set it in a new selection?

projects/components/src/checkbox/checkbox.component.ts
```diff
@@ -12,8 +12,8 @@ import { MatCheckboxChange } from '@angular/material/checkbox';
       (change)="onCheckboxChange($event)"
       [disabled]="this.disabled"
       class="htc-checkbox"
-      (click)="$event.stopPropagation()"
-      ><htc-label class="label" *ngIf="this.label" [label]="this.label"></htc-label>
+    >
```
aaron-steinfeld, line 15: This concerns me a bit - have we looked through the usages?

skjindal93: right. I will do that or make it configurable. Just trying to get the table select working first

projects/components/src/table/table.component.ts
```diff
@@ -118,7 +118,7 @@ import {
         *cdkRowDef="let row; columns: this.visibleColumns()"
         (mouseenter)="this.onDataRowMouseEnter(row)"
         (mouseleave)="this.onDataRowMouseLeave()"
-        [ngClass]="{ 'selected-row': this.isSelectedRow(row), 'hovered-row': this.isHoveredRow(row) }"
+        [ngClass]="{ 'selected-row': this.isASelectedRow(row), 'hovered-row': this.isHoveredRow(row) }"
```
aaron-steinfeld, line 121: `isRowSelected` would read better

projects/components/src/table/table.component.ts
```diff
@@ -174,9 +174,7 @@ export class TableComponent
     width: '32px',
     visible: true,
     renderer: StandardTableCellRendererType.Checkbox,
-    onClick: (row: StatefulTableRow) => {
-      this.toggleRowSelection(row);
-    }
+    onClick: (row: StatefulTableRow) => this.toggleRowSelected(row)
```
aaron-steinfeld, line 177: Is this being used?

projects/components/src/table/table.component.ts
```diff
@@ -407,15 +408,11 @@ export class TableComponent
...
-
+  public toggleRowSelected(row: StatefulTableRow): void {
     const rowSelections = this.selections ?? [];
-    this.selections = rowSelections.includes(row)
-      ? rowSelections.filter(selection => selection !== row)
-      : rowSelections.concat(row);
-    this.selectionsChange.emit(this.selections);
-
+    const rowIndexInSelections = rowSelections.findIndex(selection => isEqualIgnoreFunctions(selection, row));
+    rowIndexInSelections >= 0 ? rowSelections.splice(rowIndexInSelections, 1) : rowSelections.push(row);
```
aaron-steinfeld, line 414: why do we need to mutate this? This is an external reference that we're mutating, even though we hid it by reassigning it.

projects/components/src/table/table.component.ts
```diff
@@ -446,14 +443,11 @@ export class TableComponent
...
   }
 
-  public isCheckboxColumn(columnConfig: TableColumnConfig): boolean {
-    return columnConfig === this.multiSelectRowColumnConfig;
-  }
-
-  public isSelectedRow(row: StatefulTableRow): boolean {
+  public isASelectedRow(row: StatefulTableRow): boolean {
     return (
+      this.selectionMode !== TableSelectionMode.Multiple &&
```
aaron-steinfeld, line 448: a row can't be selected in multiselect mode?

projects/components/src/table/table.component.ts
```diff
@@ -278,6 +276,15 @@ export class TableComponent
       this.columnConfigsSubject.next(this.buildColumnConfigs());
     }
 
+    if (changes.selections) {
+      this.dataSource?.unselectAllRows();
+      this.selections?.forEach(row => {
```
aaron-steinfeld, line 281: What's the reasoning for unselecting all rows and explicitly setting them again, rather than modifying only the changed rows? It's not clear to me, so probably worth a comment (or code change)

projects/components/src/table/data/table-cdk-row-util.ts
```diff
@@ -151,6 +151,20 @@ export namespace TableCdkRowUtil {
     return [row, ...children];
   };
 
+  export const selectAllRows = (rows: StatefulTableRow[]): StatefulTableRow[] =>
```
aaron-steinfeld, line 154: It looks like we only ever use unselect. If selectAll isn't used, can we remove it? Or did I miss where it was used

projects/components/src/table/table.component.ts
skjindal93: Yes. The checkbox click actually triggers this

projects/components/src/table/table.component.ts
skjindal93: Only the checkbox should display checked in the multi select mode as per the mocks, because a checked checkbox is already an indication that the row is selected

projects/components/src/table/table.component.ts
skjindal93: We need to unselect the rows which were already checked before, if there is a change in the selections

That can either be done by maintaining a state of selected rows, or unselecting everything before and then selecting the newly selected ones

The current piece of code looks cleaner to me, instead of maintaining a state of selected rows

projects/components/src/table/data/table-cdk-row-util.ts
skjindal93: right. removing it

projects/components/src/table/table.component.ts
skjindal93: Either we have to keep a local state to track the state of selections, or mutate the selections reference.

Mutating an Input is not good, because it would indirectly trigger `ngOnChanges`?

projects/components/src/table/table.component.ts
aaron-steinfeld: ok the `ifCheckbox` escapehatch was removed, got it

projects/components/src/table/table.component.ts
aaron-steinfeld: would `changes.selections?.previousValue` suffice?

projects/components/src/table/table.component.ts
aaron-steinfeld: nope (even if we could/did this outside, onChanges is always based on references). Mutations aren't good because it makes it very difficult to understand the side effects of a change and very bug prone code - this table change feels like an example of that to me. A little more concretely, imagine I'm doing something (outside table) like assigning the array of selected rows to another component to render (say, sankeys). When that reference is mutated, the internals of the table potentially just modified a sibling component (or a parent, in other scenarios). Or maybe the reference was mutated and then emitted through the output - but since the reference is still the same, it doesn't trigger change detection in the sibling.

tl;dr: Angular is built on the idea of unidirectional data flow, and mutation breaks that.

projects/components/src/table/table.component.ts
aaron-steinfeld: It sound like we're conflating if a row is selected vs if we should style a row in the single select style. A row is selected independent of mode. Maybe call this `shouldRowHighlightAsSelection` or something?

projects/components/src/table/table.component.ts
skjindal93: I tried maintaining a local state to achieve this, but couldn't make it working

skjindal93: ignore the above comment. fixed non mutating the Input selections

projects/components/src/table/table.component.ts
skjindal93: Tried that. But, there is something which needs to go through data source using `cachedRows`. Couldn't seem to work around that

projects/components/src/table/table.component.ts
```diff
@@ -410,10 +412,9 @@ export class TableComponent
   }
 
   public toggleRowSelected(row: StatefulTableRow): void {
-    const rowSelections = [...(this.selections ?? [])];
-    const rowIndexInSelections = rowSelections.findIndex(selection => isEqualIgnoreFunctions(selection, row));
-    rowIndexInSelections >= 0 ? rowSelections.splice(rowIndexInSelections, 1) : rowSelections.push(row);
-    this.selectionsChange.emit(rowSelections);
+    const rowIndexInSelections = this.rowSelections.findIndex(selection => isEqualIgnoreFunctions(selection, row));
+    rowIndexInSelections >= 0 ? this.rowSelections.splice(rowIndexInSelections, 1) : this.rowSelections.push(row);
```
aaron-steinfeld, line 416: nit: since this isn't mutating any more, concat/without would be more readable

PR #62, opened 2020-08-06
refactor: move filter bar to components

jake-bassett:


projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.test.ts
```diff
@@ -13,16 +19,55 @@ import { ApplicationAwareDashboardComponent } from './application-aware-dashboar
 import { NavigableDashboardComponent } from './navigable-dashboard.component';
 
 describe('Navigable dashboard component', () => {
+  const attributes: FilterAttribute[] = [
```
aaron-steinfeld, line 22: These aren't used by the test are they? If not, let's just provide an empty array. The more we add things like this to test, the larger the impact of a change to `FilterAttribute`

projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.ts
```diff
@@ -67,6 +72,14 @@ export class NavigableDashboardComponent implements OnChanges {
         defaultIfEmpty(this.defaultJson)
       );
     }
+
+    if (changeObject.filterConfig && this.hasFilterBarConfig()) {
+      this.attributes$ = this.metadataService.getFilterAttributes(this.filterConfig!.filterBar!.scope);
```
aaron-steinfeld, line 77: Pretend `!` doesn't exist, pretty much always. If the compiler can't figure it out, it generally means we could do a better job typing (with a few exceptions like map and array access). Here, one issue would be that if I were to set filterConfig to undefined, we'd leave `attributes$` set to something incorrect. Instead, we could write this as something like
```
    if (changeObject.filterConfig) {
      this.attributes$ = this.filterConfig?.filterBar ? this.metadataService.getFilterAttributes(this.filterConfig.filterBar.scope) : of([]);
    }
```

projects/distributed-tracing/src/shared/services/filter-builder/graphql-filter-builder.service.ts
```diff
@@ -14,7 +14,7 @@ export class GraphQlFilterBuilderService {
         new GraphQlFieldFilter(
           filter.field,
           this.toGraphQlOperator(filter.operator),
-          this.toGraphQlArgumentValue(filter.metadata.type, filter.value)
+          this.toGraphQlArgumentValue(filter.metadata.type as AttributeMetadataType, filter.value)
```
aaron-steinfeld, line 17: this is that cast back - if we correctly type FilterAttribute, then no cast should be required

projects/components/src/filter-bar/filter-bar.component.ts
```diff
@@ -95,32 +94,22 @@ export class FilterBarComponent implements OnChanges {
   ) {}
 
   public ngOnChanges(changes: TypedSimpleChanges<this>): void {
-    if (changes.scope) {
+    if (changes.attributes) {
       this.onFiltersChanged(this.filters || []);
-      this.subscribeToUrlFilters();
+      this.readFromUrlFilters();
```
aaron-steinfeld, line 99: although existing logic, it feels like there's a lot of issues with this ngOnChanges block. It probably wasn't noticed because we don't generally change filter bar inputs. Line 98 for example - why would we emit a filter change when attributes have changed? readFromUrlFilters then calls onFiltersChanged on the next line, and may be called a second time on 103. onFiltersChanged is also called on 107, potentially emitting four times for a single change.

projects/components/src/filter-bar/filter-bar.component.ts
```diff
@@ -129,12 +118,20 @@ export class FilterBarComponent implements OnChanges {
...
 
+    if (this.syncWithUrl && this.hasAttributes()) {
+      this.writeToUrlFilter();
+    }
+  }
+
+  private readFromUrlFilters(): void {
     if (this.syncWithUrl) {
-      this.publishToUrlFilter();
+      this.onFiltersChanged(this.filterBarService.getUrlFilters(this.attributes || []));
```
aaron-steinfeld, line 128: Read should only be called if syncWithUrl is true (which is what we do for write, so they behave differently)

projects/components/src/filter-bar/filter-bar.component.ts
```diff
@@ -93,42 +93,31 @@ export class FilterBarComponent implements OnChanges {
...
-      this.readFromUrlFilters();
+  public ngOnChanges(): void {
+    if (!!this.attributes) {
+      this.syncWithUrl ? this.readFromUrlFilters() : this.onFiltersChanged(this.filters || [], false);
     }
+  }
 
-    if (changes.syncWithUrl && this.hasAttributes()) {
+  public ngOnInit(): void {
+    if (this.syncWithUrl && !!this.attributes) {
```
aaron-steinfeld, line 103: Wouldn't this case be covered by ngOnChanges?

projects/distributed-tracing/src/shared/dashboard/dashboard-wrapper/navigable-dashboard.component.test.ts
aaron-steinfeld: This comment still applies I believe

jake-bassett: Did you refresh? Looks updated to me.

PR #64, opened 2020-08-06
feat: exposing new properties to navigation

anandtiwary:
- Added resizable input property to control whether navigation can resize or not
- collapsed state is added as an input to control the view
- output property propogates collapsedChange propogates changes to property
  from inside the component
- Moved preference service deps to parent component
- removed custom webhook icon.


projects/components/src/navigation/navigation-list.component.scss
```diff
@@ -8,24 +8,23 @@
 
   height: 100%;
   width: 100%;
-  padding: 24px 5px 5px 8px;
```
anandtiwary, line 11: Padding will be provided in parent component. Right padding is required for resize icon

projects/components/src/navigation/navigation-list.component.ts
```diff
@@ -35,64 +35,68 @@ import { IconSize } from '../icon/icon-size';
...
   }
 
   private findActiveItem(navItems: NavItemConfig[]): NavItemLinkConfig | undefined {
     return navItems
       .filter((item): item is NavItemLinkConfig => item.type === NavItemType.Link)
       .find(linkItem =>
-        linkItem.matchPaths.some(matchPath =>
-          this.navigationService.isRelativePathActive([matchPath], this.navigationService.rootRoute())
-        )
+        this.navigationService.isRelativePathActive(linkItem.matchPaths, this.navigationService.rootRoute())
```
aaron-steinfeld, line 99: This seems like a bug - match paths should be `OR'd` (hence the previous `some`)

aaron-steinfeld:
> * Moved preference service deps to parent component

The HTC version of the left nav should be updated in this PR

projects/components/src/navigation/navigation-list.component.ts
anandtiwary: For settings, the navigation match path would be `['settings', 'users']`. `isRelativePathActive` check fails for such a case. Why do we want to `or` this?

aaron-steinfeld: that should be expressable as `['settings/users']`

anandtiwary: I see. Right not there are none and this use case was also not clear from the api. Updating it

aaron-steinfeld: 👍 I think currently the IA does not have this use case, but it comes and goes so retaining support makes sense.

PR #71, opened 2020-08-10
feat(table): added page inputs

skjindal93:


skjindal93:
Looking for first round of reviews, before adding unit test

projects/components/src/table/table.component.ts
```diff
@@ -234,6 +241,9 @@ export class TableComponent
   @Output()
   public readonly toggleAllChange: EventEmitter<boolean> = new EventEmitter<boolean>(); // True: expand, False: collapse
 
+  @Output()
+  public readonly pageChange: EventEmitter<boolean> = new EventEmitter<boolean>();
```
aaron-steinfeld, line 245: Why a boolean? I'd think this would emit the new page index - corresponding to `page` the input, which we're not currently exposing.

projects/components/src/table/table.component.ts
```diff
@@ -517,7 +528,7 @@ export class TableComponent
   private pageDataFromUrl(params: ParamMap): Partial<PageEvent> | undefined {
```
aaron-steinfeld, line 528: This (and the observable) are no longer just url based, so should be renamed.

projects/components/src/table/table.component.ts
```diff
@@ -517,7 +528,7 @@ export class TableComponent
...
     return this.syncWithUrl
       ? {
-          pageSize: new NumberCoercer().coerce(params.get(TableComponent.PAGE_SIZE_URL_PARAM)),
+          pageSize: new NumberCoercer().coerce(params.get(TableComponent.PAGE_SIZE_URL_PARAM)) ?? this.pageSize,
```
aaron-steinfeld, line 531: Same effect, but slightly more readable IMO if we declare defaultValue explicitly. 
```
pageSize: new NumberCoercer({defaultvalue: this.pageSize}).coerce(params.get(TableComponent.PAGE_SIZE_URL_PARAM))
```

projects/components/src/table/table.component.ts
skjindal93: makes sense. emitting the entire page event now

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -214,6 +214,7 @@ export class TableCdkDataSource implements DataSource<TableRow> {
       map(response => response.data),
       map(rows => this.paginateRows(rows, pageEvent)),
       map(TableCdkRowUtil.buildInitialRowStates),
+      map(TableCdkRowUtil.unselectAllRows),
```
aaron-steinfeld, line 217: Why is this needed? Wouldn't this clear initial selections?

skjindal93: I added that because rows were getting randomly selected while navigating through multiple table pages through paginator. 

But, it seems to be fixed now after @jake-bassett's latest fix for multi select on the table

PR #72, opened 2020-08-10
feat: multi select component

arjunlalb:


projects/components/src/select/select.component.ts
```diff
@@ -32,28 +32,38 @@ import { SelectSize } from './select-size';
...
+      <htc-popover [disabled]="this.disabled" [closeOnClick]="!this.multiSelectMode" class="select-container">
         <htc-popover-trigger>
           <div class="trigger-content" [ngClass]="this.justifyClass">
             <htc-icon *ngIf="this.icon" class="trigger-prefix-icon" [icon]="this.icon" size="${IconSize.Small}">
             </htc-icon>
-            <htc-label class="trigger-label" [label]="selected?.label || this.placeholder"> </htc-label>
+            <htc-label
+              *ngIf="!this.multiSelectMode"
+              class="trigger-label"
+              [label]="selected?.label || this.placeholder"
```
aaron-steinfeld, line 49: No real reason for this if/else right? the create trigger label logic supports both modes. It might be good to do it through na observable or save it too a var so we don't have to call this function each cycle.

projects/components/src/select/select.component.ts
```diff
@@ -90,13 +100,16 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
...
+  public multiSelectMode: boolean = false;
+
   @Output()
   public readonly selectedChange: EventEmitter<V> = new EventEmitter<V>();
 
   @ContentChildren(SelectOptionComponent)
   public items?: QueryList<SelectOptionComponent<V>>;
 
-  public selected$?: Observable<SelectOption<V> | undefined>;
+  public selected$?: Observable<SelectOption<V> | SelectOption<V>[] | undefined>;
```
aaron-steinfeld, line 112: Cleaner to always work in arrays IMO

projects/components/src/select/select.component.ts
```diff
@@ -90,13 +100,16 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
   @Input()
   public highlightSelected: boolean = true;
 
+  @Input()
+  public multiSelectMode: boolean = false;
```
aaron-steinfeld, line 104: All of the inputs/outputs need to support arrays now, right? or unlink them, support both in, and just emit the selection change each time (but would require metadata saying if it was selected or unselected)

projects/components/src/select/select.component.ts
```diff
@@ -144,18 +176,39 @@ export class SelectComponent<V> implements AfterContentInit, OnChanges {
   }
 
   public onSelectionChange(item: SelectOptionComponent<V>): void {
-    this.selected = item.value;
+    if (this.multiSelectMode) {
+      const selectionArray = (this.selected ?? []) as SelectOption<V>[];
+      if (selectionArray.map(option => option.value).includes(item.value)) {
+        ((this.selected ?? []) as SelectOption<V>[]).forEach((option, index) => {
+          if (option.value === item.value) {
+            ((this.selected ?? []) as SelectOption<V>[]).splice(index, 1);
```
aaron-steinfeld, line 184: Shouldn't be mutating array.

projects/components/src/select/select.component.ts
arjunlalb: yea. no reason for if/else. I was only testing. will be cleaning this up

projects/components/src/select/select.component.ts
arjunlalb: It'll definitely make code a lot more cleaner.

projects/components/src/select/select.component.ts
arjunlalb: This is the reason why I was working with an OR'ed type instead of taking in arrays. Any usages of the emitted events in all consumers have to be modified. 
In that case, do you think a separate component would be better? @aaron-steinfeld

aaron-steinfeld: Given how similar they are, I'd lean away from a separate components.
So we could:
1. update all the consumers (hopefully not too many).
2. Have a union of types on inputs + outputs (in this model, I'd at least convert to arrays internally to keep it clean there. Could also make the union a generic type, so the caller can narrow it based on their use case.
3. Have different outputs (and potentially inputs) for different modes. (this strikes me as bad)

I'd probably lean towards 2 (iterating on what we have here)
What are other thoughts @anandtiwary @jake-bassett ?

arjunlalb: Number 2 was what I was aiming for. But after making these changes and seeing whats ahead, I think a different component fits better.
In case of `selectedChange$` for single select, the change always has one value. In case of multi-select, this contract changes. The consumers will have to take the burden of casting appropriately and consuming the values.
Similarly, some styling changes also does not apply for multi select. 

More I think of this, I feel like its cleaner to have a separate multi select component. I don't think we gain much by trying to reuse code with union types. Code (both template and TS) will be messier with more casting / if else blocks.

projects/components/src/select/select.component.ts
arjunlalb: I think it makes sense to use union types so that existing consumers don't need to be modified.

projects/components/src/select/select.component.ts
aaron-steinfeld: Sure, let's go with the separate components then.

jake-bassett: +1 to separate components

projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -0,0 +1,91 @@
+@import 'mixins';
+
+:host {
```
aaron-steinfeld, line 3: Could we do this without using host? A div inside? This prevents a user from setting a width for their own use case. Discussed here https://github.com/hypertrace/hypertrace-core-ui/pull/57#discussion_r464719695

projects/components/src/multi-select/multi-select.component.scss
```diff
@@ -0,0 +1,91 @@
...
+  width: 100%;
+  max-width: 240px;
+
+  .multi-select {
```
aaron-steinfeld, line 7: any reason this is nested inside the host block rather than inside the `.multi-select` block below?

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,164 @@
...
+  public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (this.items !== undefined && changes.selected !== undefined) {
+      this.selected$ = this.buildObservableOfSelected();
+    }
+    this.setTriggerLabel();
+  }
+
+  private setTriggerLabel(): void {
+    const selectedItems: SelectOptionComponent<V>[] | undefined = this.items?.filter(
+      item => this.selected !== undefined && this.selected?.includes(item.value)
```
aaron-steinfeld, line 109: could this reuse `isSelectedItem`?

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,164 @@
...
+        this.selected.push(item.value);
+      }
+    }
+    this.setTriggerLabel();
+    this.selected$ = this.buildObservableOfSelected();
+    this.selectedChange.emit(this.selected);
+  }
+
+  // Find the select option object for a value
+  private findItem(value: V[] | undefined): SelectOption<V>[] | undefined {
```
aaron-steinfeld, line 155: pluralize

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    return queryListAndChanges$(this.items).pipe(
+      switchMap(items => merge(of(undefined), ...items.map(option => option.optionChange$))),
+      map(() => this.findItem(this.selected))
+    );
+  }
+
+  public onSelectionChange(item: SelectOptionComponent<V>): void {
+    // If selected in undefined
+    if (this.selected === undefined) {
```
aaron-steinfeld, line 137: this branching could be trivially removed by changing line 140 to account for the undefined case.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,164 @@
...
+      this.selected = [item.value];
+    } else {
+      const selectedItems: V[] = [...this.selected];
+      if (selectedItems.includes(item.value)) {
+        // Remove if already selected
+        this.selected = selectedItems.filter(value => value !== item.value);
```
aaron-steinfeld, line 143: feels like we have 3 or 4 places in this component where we are doing a filter/include type check to map between selections and items - we should be able to define that in one spot.

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -0,0 +1,164 @@
...
+import { SelectJustify } from '../select/select-justify';
+import { SelectOption } from '../select/select-option';
+import { SelectOptionComponent } from '../select/select-option.component';
+import { SelectSize } from '../select/select-size';
+
+@Component({
+  selector: 'htc-multi-select',
+  styleUrls: ['./multi-select.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  providers: [SubscriptionLifecycle],
```
aaron-steinfeld, line 26: unused

projects/components/src/multi-select/multi-select.component.scss
arjunlalb: I copied the base component from the select component. working on cleaning up things now. We don't need the host here. Its left over from using material components inside select I think.

projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -0,0 +1,162 @@
...
+import { fakeAsync, flush } from '@angular/core/testing';
+import { NavigationService } from '@hypertrace/common';
+import { MultiSelectComponent, MultiSelectModule, SelectJustify, SelectModule } from '@hypertrace/components';
+import { createHostFactory, mockProvider, SpectatorHost } from '@ngneat/spectator/jest';
+import { EMPTY } from 'rxjs';
+
+describe('Multi Select Component', () => {
+  const hostFactory = createHostFactory<MultiSelectComponent<string>>({
+    component: MultiSelectComponent,
+    imports: [MultiSelectModule, SelectModule, HttpClientTestingModule],
```
aaron-steinfeld, line 11: Is selectmodule used? Also can this be made shallow (may not be worthwhile due to popover, up to you - or maybe shallow with just the popovermodule import)

projects/components/src/multi-select/multi-select.module.ts
```diff
@@ -0,0 +1,16 @@
...
+import { FormsModule } from '@angular/forms';
+import { IconModule } from '../icon/icon.module';
+import { LabelModule } from '../label/label.module';
+import { LetAsyncModule } from '../let-async/let-async.module';
+import { PopoverModule } from '../popover/popover.module';
+import { SelectModule } from '../select/select.module';
+import { MultiSelectComponent } from './multi-select.component';
+
+@NgModule({
+  imports: [FormsModule, CommonModule, IconModule, LabelModule, LetAsyncModule, PopoverModule, SelectModule],
```
aaron-steinfeld, line 12: SelectModule used?

projects/components/src/multi-select/multi-select.component.ts
```diff
@@ -130,13 +130,9 @@ export class MultiSelectComponent<V> implements AfterContentInit, OnChanges {
...
   public onSelectionChange(item: SelectOptionComponent<V>): void {
-    if (this.isSelectedItem(item)) {
-      // Remove if already selected
-      this.selected = this.selected?.filter(value => value !== item.value);
-    } else {
-      // Add if not present already
-      this.selected = (this.selected ?? []).concat(item.value);
-    }
+    this.selected = this.isSelectedItem(item)
+      ? (this.selected = this.selected?.filter(value => value !== item.value))
```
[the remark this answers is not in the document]
arjunlalb: oops. fixed it.

projects/components/src/multi-select/multi-select.component.test.ts
arjunlalb: SelectModule is needed in the test for the option component. Also, tried making this shallow. Doesn't work.

skjindal93: Made it shallow

arjunlalb: Thank you @skjindal93 ! :)

projects/components/src/multi-select/multi-select.component.test.ts
```diff
@@ -0,0 +1,162 @@
+import { fakeAsync, flush } from '@angular/core/testing';
+import {
+  LabelComponent,
+  LetAsyncModule,
+  MultiSelectComponent,
+  SelectJustify,
+  SelectOptionComponent
+} from '@hypertrace/components';
```
anandtiwary, line 8: Same lib import

PR #73, opened 2020-08-10
fix: moving away form batched queries

anandtiwary:
- Currently, we batch few queries in a dashboard and send them together in the same network call. This means that:

   -  The overall query time is governed by slowest individual query.

   - For a large time range, this cause significant delay on the dashboard load.  Most of the widgets are not visible (till the slowest query call is completed) while the user is waiting for it.

   - Once all the data arrives, browser would spend time in scripting and rendering of all queries one after the other, which adds to overall response time

If we make individual queries, then some widgets will show data sooner than others thereby giving a shorter `perceived` response time. It would also make better use of browser idle time as scripting and rendering for faster queries would be done by the time slower queries results are returned. So overall exp would be improved.


anandtiwary:
Sure. Since I am completely going away from batching, I decided to change the individual data source models which calls on queryWithNextBatch directly. 

From the widget's perspective, Do you still feel it would be useful? With my poc I came to a conclusion that batching is not good for our dashboards.

aaron-steinfeld:
Discussed offline - we'll iterate on how to batch or not, this change will make it so it's opaque to the widgets

PR #75, opened 2020-08-11
feat: filter button component in explorer

jake-bassett:


projects/components/src/table/renderers/table-cell-renderer.component.ts
```diff
@@ -12,31 +11,59 @@ import {
...
+  public initialize(): void {
+    this._value = this.parseValue(this.cellData, this.rowData);
+    this._units = this.parseUnits(this.cellData, this.rowData);
+    this._tooltip = this.parseTooltip(this.cellData, this.rowData);
+  }
+
+  protected abstract parseValue(cellData: TCellData, rowData?: TableRow): TValue;
+
+  protected parseUnits(_cellData: TCellData, _rowData?: TableRow): string {
+    return '';
```
arjunlalb, line 47: is this intentionally left empty?

jake-bassett: Yep. Many attributes don't have units, so this is sort of a no-op to build the default tooltip.

projects/components/src/filter-bar/filter-bar.service.ts
```diff
@@ -1,17 +1,27 @@
...
 import { Filter } from './filter/filter-api';
 import { FilterParserService } from './filter/parser/filter-parser.service';
 
+export type GetFiltersFunction = (attributes: FilterAttribute[]) => Filter[];
+
 @Injectable({ providedIn: 'root' })
 export class FilterBarService {
   private static readonly FILTER_QUERY_PARAM: string = 'filter';
 
+  public readonly urlFilterChanges$: Observable<GetFiltersFunction>;
```
aaron-steinfeld, line 15: IMO it would be cleaner to have a function taking an array of attributes and returning an observable<filter[]>. Thoughts?

projects/components/src/table/renderers/table-cell-renderer.component.ts
```diff
@@ -12,31 +11,59 @@ import {
...
   public readonly isFirstColumn: boolean = false;
 
-  protected abstract parseValue(cellData?: TRaw, rowData?: Dictionary<unknown>): TParsed;
-
   public constructor(
     @Inject(TABLE_CELL_RENDERER_COLUMN_CONFIG) private readonly columnConfig: TableColumnConfig,
     @Inject(TABLE_CELL_RENDERER_COLUMN_INDEX) private readonly index: number,
-    @Optional() @Inject(TABLE_CELL_RENDERER_CELL_DATA) private readonly cellData: TRaw | null,
-    @Optional() @Inject(TABLE_CELL_RENDERER_ROW_DATA) private readonly rowData: Dictionary<unknown> | null
+    @Inject(TABLE_CELL_RENDERER_ROW_DATA) private readonly rowData: TableRow,
```
aaron-steinfeld, line 27: Do you remember why these were optional? and why was it changed?

projects/components/src/table/renderers/table-cell-renderer.component.ts
```diff
@@ -12,31 +11,59 @@ import {
...
     this.clickable = this.columnConfig.onClick !== undefined;
     this.isFirstColumn = this.index === 0;
   }
 
   public ngOnInit(): void {
-    this._value = this.parseValue(this.cellData ?? undefined, this.rowData ?? undefined);
+    this.initialize();
+  }
+
+  public initialize(): void {
```
aaron-steinfeld, line 38: why a separate (public) initialize function called by init? can we inline?

projects/components/src/table/table.component.ts
```diff
@@ -185,6 +188,9 @@ export class TableComponent
   @Input()
   public columnConfigs?: TableColumnConfig[];
 
+  @Input()
+  public metadata?: FilterAttribute[];
```
aaron-steinfeld, line 192: Strikes me as strange that the table needs this. I guess let me see what data cell renderer is doing with it.

projects/components/src/filter-button/filter-button.service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  public buildAvailableFilters(attribute: FilterAttribute, value: unknown): Filter[] {
+    return this.filterBuilderService.lookup(attribute).buildFiltersForAvailableOperators(attribute, value);
+  }
+
+  /*
+   * TODO: Combine service with filter-bar.service. A few duplicated methods here.
+   */
+
+  public applyUrlFilter(attributes: FilterAttribute[], filter: Filter): void {
```
aaron-steinfeld, line 26: If we can remove attributes from this API (which makes sense I would think, we don't need attributes to convert to a url param) then I think we can clean up a bunch - including tying attributes into the table directly. So as I read this we need these to read in the current filters, replace if a field match, and otherwise append. But we could probably write some logic to do the same without doing a full parse (e.g. split on any operator and convert to a KV map, where K is the source name, V is the original url string)

projects/components/src/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  @Output()
+  public readonly popoverOpen: EventEmitter<boolean> = new EventEmitter();
+
+  public availableFilters: Filter[] = [];
+
+  public constructor(private readonly filterButtonService: FilterButtonService) {}
+
+  public ngOnChanges(): void {
+    if (this.attribute !== undefined) {
```
aaron-steinfeld, line 60: if it is undefined, we should clear the available filters

projects/components/src/filter-button/filter-button.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  public metadata?: FilterAttribute[];
+
+  @Input()
+  public attribute?: FilterAttribute;
+
+  @Input()
+  public value?: T;
+
+  @Output()
+  public readonly apply: EventEmitter<Filter> = new EventEmitter();
```
aaron-steinfeld, line 50: Is this output used? because it doesn't look like we emit on it.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
```diff
@@ -1,54 +1,95 @@
...
-      <div class="horizontal-alignment-container" [ngClass]="this.alignment">
+    <div class="table-data-cell-renderer" [class.selected]="this.popoverOpen">
+      <htc-filter-button
+        class="filter-button"
+        *ngIf="this.columnConfig?.filterable && this.leftAlignFilterButton"
+        [metadata]="this.metadata"
+        [attribute]="this.columnConfig?.attribute"
+        [value]="this.componentRef?.instance.value"
+        (popoverOpen)="this.popoverOpen = $event"
+      ></htc-filter-button>
```
aaron-steinfeld, line 30: there's enough in this that it might be worth using a template rather than repeating it for either alignment

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
```diff
@@ -1,54 +1,95 @@
...
   template: `
-    <div class="vertical-alignment-container">
-      <div class="horizontal-alignment-container" [ngClass]="this.alignment">
+    <div class="table-data-cell-renderer" [class.selected]="this.popoverOpen">
+      <htc-filter-button
+        class="filter-button"
+        *ngIf="this.columnConfig?.filterable && this.leftAlignFilterButton"
+        [metadata]="this.metadata"
+        [attribute]="this.columnConfig?.attribute"
+        [value]="this.componentRef?.instance.value"
```
aaron-steinfeld, line 28: what can we do to avoid reaching into a component from its parent

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
```diff
@@ -1,54 +1,95 @@
...
+    private readonly componentFactoryResolver: ComponentFactoryResolver
   ) {}
 
   public ngOnInit(): void {
     if (this.columnConfig === undefined) {
       throw new Error('Table column config undefined');
     }
 
+    if (this.columnConfig.filterable && !this.columnConfig.attribute) {
+      throw new Error(`Table attribute required to enable filtering for field '${this.columnConfig.field}'`);
```
aaron-steinfeld, line 81: this goes to my other comment about the api - no need to have both configs.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
```diff
@@ -58,7 +99,14 @@ export class TableDataCellRendererComponent implements OnInit {
       this.injector
     );
 
+    // Use cell renderer to initialize value for filtering
+    this.componentRef = this.componentFactoryResolver
```
aaron-steinfeld, line 103: Oh god - this is why you added initialize. We're instantiating a component but never attaching it to a template and just manipulating it to pull value out of it... we gotta come up with a better way of doing this.

projects/components/src/filter-button/filter-button.service.ts
jake-bassett: I can see the argument from a cleanliness perspective, but I'm not keen on making the presence of an object run dual purpose and also mean we want to filter. It could make the API confusing.

I'll make this change for now, but I'm posting this so I can link to it some day in the future. :-)

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
jake-bassett: Fair. Let's see how all of the other fixes go and I can circle back to this if I have time.

projects/components/src/table/table-api.ts
```diff
@@ -1,15 +1,18 @@
...
 export interface TableColumnConfig {
   field: string;
+  attribute?: FilterAttribute;
   title?: string;
   titleTooltip?: string;
   renderer?: string;
   sort?: TableSortDirection;
   visible?: boolean;
   sortable?: boolean;
+  filterable?: boolean;
```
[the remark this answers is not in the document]
jake-bassett: Will remove.

projects/components/src/table/renderers/table-cell-renderer.component.ts
jake-bassett: I can't remember exactly why they were optional, but I suspect it might have been something with the expander renderer not going through the service originally. I fixed that, so this seems correct now.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
jake-bassett: The issues you noted in this file are the ugly parts I alluded to. Agree we need a better way. Let's have the conversation here and I'll link the other comments about the cell renderer code to this one.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
jake-bassett: Removing filterable.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
jake-bassett: https://github.com/hypertrace/hypertrace-core-ui/pull/75/files#r469530453

projects/components/src/table/renderers/table-cell-renderer.component.ts
jake-bassett: https://github.com/hypertrace/hypertrace-core-ui/pull/75/files#r469530453

projects/components/src/table/table.component.ts
jake-bassett: https://github.com/hypertrace/hypertrace-core-ui/pull/75/files#r469530453

projects/components/src/filter-button/filter-button.service.ts
aaron-steinfeld: This sounds like a response to https://github.com/hypertrace/hypertrace-core-ui/pull/75#discussion_r468958050 ?

If so, we could tweak the name from attribute to `filterAttribute` or something, but what I want to avoid is having separate inputs saying should I do something and how to do it - unless it's something that is meaningful to support changing at runtime and the like, such as column visibility. If you tell me (me being the table) how to filter, then I should expose it and if you don't tell me how, I shouldn't. The separation leads to either an unnecessary error conditions where you mark something as filterable but don't have an attribute, or extra data when you tell me how to filter but don't use it.

That said, my original comment above about removing the attributes coming in to table I think is one of the more important ones on the review so let's revisit that one.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
aaron-steinfeld: so my starting question here would actually be - do we need the renderer at all to get the value? I would think filtering generally should be based off the raw value rather than however the renderer packages it for display - although this can actually go both ways so neither way is perfectly correct. In one direction for example, take the backend icon renderer - it takes a backend type and converts it into an icon type for rendering. If we were to filter on that, we'd want to take that raw backend type. In the other direction, something like metric currently gets mapped to an object containing units, then unpacked in the renderer.

So, I guess what I'm trying to say is that we've overloaded the concept of value in renderer - the value to use for building a filter often overlaps with the raw value coming in, or the renderer produced value, but both are something different. Given that, the most appropriate place (ignoring implementation) I would think would be in the specification - that's where we define `asGraphQlOrderByFragment` which is a similar use case.

projects/components/src/filter-bar/filter-type.ts
```diff
@@ -1,7 +1,25 @@
 export const enum FilterType {
   // From AttributeMetadataType
+  Boolean = 'BOOLEAN',
   String = 'STRING',
   Number = 'LONG',
   StringMap = 'STRING_MAP',
   Timestamp = 'TIMESTAMP'
 }
+
+export const toFilterType = (type: string): FilterType => {
```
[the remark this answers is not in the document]
aaron-steinfeld: It's called from the table widget (because it has attributes coming in), which is in distributed tracing. Unless I'm missing a usage.

projects/components/src/filter-bar/filter-bar.service.ts
aaron-steinfeld: At first, you convinced me to leave it as is, then I dug in - the current implementation doesn't handle the case you described either, because the function is only invoked if `urlFilterChanges$` emits. So we'd need to do something like converting the input to an observable too and zipping them together - at which point, there's no extra management of subscriptions between the two approaches.

projects/components/src/filter-button/filter-button.service.ts
aaron-steinfeld: > As for removing attributes from table, I think that's just too much right now. All of our filtering code depends on the metadata and I think our time is better spent elsewhere than rewriting that for this PR. 

I'd agree that we shouldn't rewrite that here, but I don't think that's what would be needed to remove this. Unless I'm misunderstanding which is entirely possible, the only place this is used is in replacing an existing filter in the URL. We could almost certainly do a trivial parse (without touching the existing logic) to accomplish that which would not require attributes. More importantly - I'm not convinced that's the right behavior. If I have a filter like duration > 10, and add another filter like duration <15, those should both be applicable. Why are we trying to enforce a single filter per key?

projects/components/src/filter-bar/filter-bar.component.ts
```diff
@@ -98,31 +102,43 @@ export class FilterBarComponent implements OnChanges {
     }
   }
 
+  public ngOnInit(): void {
+    if (this.syncWithUrl) {
+      this.filterBarService.urlFilterChanges$
+        .pipe(map((getFilters: GetFiltersFunction) => getFilters(this.attributes || [])))
+        .subscribe(filters => {
```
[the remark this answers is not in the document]
aaron-steinfeld: We can't get away from a subject here I agree. What we could do though is avoid the susbcription. For example, we could merge our own subject of inputFilters with urlFilterChanges$ to create an observable of filters that is subscribed to by angular.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
aaron-steinfeld: So there's 3 actions we do with a a column/cell right now -
1. Display it. This takes in a value, and parses it to some type of visualization. It could do things like adding units, changing formats, joining extra data etc. Nothing here cares about where the data came from.
2. Sort it. This takes a column definition and converts it into a sort command. The sort command itself depends on the implementation of the data source, so we tell the data source which column was sorted as well as the direction,  and it handles how to do so (`TableDataRequest`). In converting a column to a sort, the specific implementation of column is responsible for determining how to materialize that. In our case, it's a graphql specification which could map to multiple attributes because we have some complex columns. Take status for example, where we join status the boolean, status code and status description. When we sort it, we want to just use the single status code column.
3. Now, filter it. This is almost a combination of 1+2 since it needs both an attribute and value, but IMO it's much closer to 2 because the value we need here needs to match the data source expected value, not the display value. The graphql specification was responsible for combining any number of attributes and producing a value out of them, so already has the knowledge and seems the appropriate place to do that in reverse.

projects/components/src/filter-button/filter-button.service.ts
aaron-steinfeld: Also (just to pin this convo into GH so we can revisit it later if needed) if we were to use a provider for communication rather than the URL (besides making this capability more portable to work in places we're not using a url sync'd table), the need to parse would go away and we'd also be able to avoid the attribute input.

anandtiwary:
It's a bit challenging to follow the PR and review. Would have been easier if it was a smaller PR :)

projects/components/src/filter-bar/filter-bar.service.ts
jake-bassett: So close! 

But really, good catch! Updating.

projects/components/src/filter-button/filter-button.service.ts
aaron-steinfeld: I guess I would think that being more lax with the filters would be a better starting point as it 
1. Allows more functionality (not just things like ranges, but also `!= x and != y`, or `like x and != xy`)
2. Is simpler to implement - we have code existing to enforce this right now, including this thread
3. Does allow users to create nonsensical queries, but that's always possible (e.g. I could query for status = success and statusCode = 500). I think that's the trade off for flexibility. The important thing is to make the query obvious. If we want to add certain rules to prevent knowably wrong things as an improvement, I think that can come iteratively (for example not allowing a second equality filter for the same field).

jake-bassett: Okay. I'll take care of it with the other jira.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
jake-bassett: So, what is the ask here then?

aaron-steinfeld: I'm not sure. I'd rather have the spec deal with this, but given the level of effort, I think the only realistic ask is to clean up how we interact with the renderer. So my compromise ask is to not create two copies of every renderer but use the one actual rendering to get the value, rather than creating and discarding a new component.

jake-bassett: That's reasonable. Let me take a look on getting that working.

projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
```diff
@@ -29,7 +30,7 @@ import { TableCellRendererService } from '../table-cell-renderer.service';
         (popoverOpen)="this.popoverOpen = $event"
       ></htc-filter-button>
       <div class="cell-renderer-content" [ngClass]="this.alignment" (click)="this.onClick()">
-        <ng-container *ngComponentOutlet="this.rendererConstructor; injector: this.rendererInjector"></ng-container>
+        <div #cellRenderer></div>
```
aaron-steinfeld, line 33: nit - I don't think this extra div is needed, if you move #cellRenderer to 32 it should be the equivalent dom to before

jake-bassett: Wasn't working before due to some css issues, but I'll see if I can adjust the styling and remove the div.

PR #76, opened 2020-08-12
feat: adding click handler that will open template in sheet

anandtiwary:
- We can apply this interaction handler to a table column.
- The row clicked will be available as the variable 'record'. This can be referenced
  in the template itself

Example usage:

```
      {
        type: 'table-widget-column',
        title: 'URL',
        value: {
          type: 'attribute-specification',
          attribute: 'requestUrl'
        },
        'click-handler': {
          type: 'detail-sheet-interaction-handler',
          'sheet-size': SheetSize.Large,
          'detail-template': {
            type: 'trace-detail-widget',
            data: {
              type: 'trace-detail-extended-data-source',
              // tslint:disable-next-line: no-invalid-template-strings
              trace: '${record}'
            }
          }
        }
      },
```

### DIFF
--- a/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
@@ -1,0 +1,12 @@
+import { ChangeDetectionStrategy, Component, Inject, InjectionToken } from '@angular/core';
+
+@Component({
+  selector: 'htc-detail-sheet-interaction-container',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<ng-container [hdaDashboardModel]="this.detailModel"></ng-container>`
+})
+export class DetailSheetInteractionContainerComponent {
+  public constructor(@Inject(DETAIL_SHEET_INTERACTION_MODEL) public readonly detailModel: object) {}
+}
+
+export const DETAIL_SHEET_INTERACTION_MODEL = new InjectionToken<unknown>('DETAIL_SHEET_INTERACTION_MODEL');
--- a/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
@@ -9,4 +9,4 @@ export class DetailSheetInteractionContainerComponent {
   public constructor(@Inject(DETAIL_SHEET_INTERACTION_MODEL) public readonly detailModel: object) {}
 }
 
-export const DETAIL_SHEET_INTERACTION_MODEL = new InjectionToken<unknown>('DETAIL_SHEET_INTERACTION_MODEL');
+export const DETAIL_SHEET_INTERACTION_MODEL = new InjectionToken<object>('DETAIL_SHEET_INTERACTION_MODEL');
--- a/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
@@ -1,6 +1,6 @@
 import { SheetSize } from '@hypertrace/components';
 import { EnumPropertyTypeInstance, ENUM_TYPE, ModelTemplatePropertyType } from '@hypertrace/dashboards';
-import { Model, ModelApi, ModelJson, ModelProperty } from '@hypertrace/hyperdash';
+import { Model, ModelApi, ModelJson, ModelProperty, STRING_PROPERTY } from '@hypertrace/hyperdash';
 import { ModelInject, MODEL_API } from '@hypertrace/hyperdash-angular';
 import { Observable, of } from 'rxjs';
 import { InteractionHandler } from '../interaction-handler';
@@ -18,14 +18,20 @@ export class DetailSheetInteractionHandlerModel implements InteractionHandler {
 
   @ModelProperty({
     key: 'sheet-size',
-    required: true,
+    required: false,
     // tslint:disable-next-line: no-object-literal-type-assertion
     type: {
       key: ENUM_TYPE.type,
       values: [SheetSize.Small, SheetSize.Medium, SheetSize.Large]
     } as EnumPropertyTypeInstance
   })
-  public sheetSize!: SheetSize;
+  public sheetSize: SheetSize = SheetSize.Large;
+
+  @ModelProperty({
+    key: 'inject-source-as',
+    type: STRING_PROPERTY.type
+  })
+  public injectSourceAs: string = 'source';
 
   @ModelInject(MODEL_API)
   private readonly api!: ModelApi;
@@ -39,9 +45,9 @@ export class DetailSheetInteractionHandlerModel implements InteractionHandler {
     return of();
   }
 
-  private getDetailModel(record: unknown): object {
+  private getDetailModel(source: unknown): object {
     const detailModel = this.api.createChild<object>(this.detailTemplate, this);
-    this.api.setVariable('record', record, detailModel);
+    this.api.setVariable(this.injectSourceAs, source, detailModel);
 
     return detailModel;
   }
--- a/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
+++ b/projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
@@ -1,0 +1,48 @@
+import { SheetSize } from '@hypertrace/components';
+import { EnumPropertyTypeInstance, ENUM_TYPE, ModelTemplatePropertyType } from '@hypertrace/dashboards';
+import { Model, ModelApi, ModelJson, ModelProperty } from '@hypertrace/hyperdash';
+import { ModelInject, MODEL_API } from '@hypertrace/hyperdash-angular';
+import { Observable, of } from 'rxjs';
+import { InteractionHandler } from '../interaction-handler';
+import { DetailSheetInteractionHandlerService } from './detail-sheet-interaction-handler.service';
+
+@Model({
+  type: 'detail-sheet-interaction-handler'
+})
+export class DetailSheetInteractionHandlerModel implements InteractionHandler {
+  @ModelProperty({
+    key: 'detail-template',
+    type: ModelTemplatePropertyType.TYPE
+  })
+  public detailTemplate!: ModelJson;
+
+  @ModelProperty({
+    key: 'sheet-size',
+    required: true,
+    // tslint:disable-next-line: no-object-literal-type-assertion
+    type: {
+      key: ENUM_TYPE.type,
+      values: [SheetSize.Small, SheetSize.Medium, SheetSize.Large]
+    } as EnumPropertyTypeInstance
+  })
+  public sheetSize!: SheetSize;
+
+  @ModelInject(MODEL_API)
+  private readonly api!: ModelApi;
+
+  @ModelInject(DetailSheetInteractionHandlerService)
+  private readonly handlerService!: DetailSheetInteractionHandlerService;
+
+  public execute(record: unknown): Observable<void> {
+    this.handlerService.showSheet(this.getDetailModel(record), this.sheetSize);
+
+    return of();
+  }
+
+  private getDetailModel(record: unknown): object {
+    const detailModel = this.api.createChild<object>(this.detailTemplate, this);
+    this.api.setVariable('record', record, detailModel);
+
+    return detailModel;
+  }
+}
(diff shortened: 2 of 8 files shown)


projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Observable, of } from 'rxjs';
+import { InteractionHandler } from '../interaction-handler';
+import { DetailSheetInteractionHandlerService } from './detail-sheet-interaction-handler.service';
+
+@Model({
+  type: 'detail-sheet-interaction-handler'
+})
+export class DetailSheetInteractionHandlerModel implements InteractionHandler {
+  @ModelProperty({
+    key: 'detail-template',
```
arjunlalb, line 14: naming comment : 
can we use a more generic name here?
This interaction handler could be used for non detail views as well. 

Perhaps just `OverlaySheetInteractionHandlerModel`, `sheet-template` etc?

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
```diff
@@ -0,0 +1,12 @@
...
+@Component({
+  selector: 'htc-detail-sheet-interaction-container',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<ng-container [hdaDashboardModel]="this.detailModel"></ng-container>`
+})
+export class DetailSheetInteractionContainerComponent {
+  public constructor(@Inject(DETAIL_SHEET_INTERACTION_MODEL) public readonly detailModel: object) {}
+}
+
+export const DETAIL_SHEET_INTERACTION_MODEL = new InjectionToken<unknown>('DETAIL_SHEET_INTERACTION_MODEL');
```
aaron-steinfeld, line 12: object, right? should match the type on line 9

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
```diff
@@ -0,0 +1,12 @@
+import { ChangeDetectionStrategy, Component, Inject, InjectionToken } from '@angular/core';
+
+@Component({
+  selector: 'htc-detail-sheet-interaction-container',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<ng-container [hdaDashboardModel]="this.detailModel"></ng-container>`
```
aaron-steinfeld, line 6: ooh haven't seen the attribute directive used directly in so long!

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
```diff
@@ -0,0 +1,48 @@
...
+    type: ModelTemplatePropertyType.TYPE
+  })
+  public detailTemplate!: ModelJson;
+
+  @ModelProperty({
+    key: 'sheet-size',
+    required: true,
```
aaron-steinfeld, line 21: should be able to default this

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  public execute(record: unknown): Observable<void> {
+    this.handlerService.showSheet(this.getDetailModel(record), this.sheetSize);
+
+    return of();
+  }
+
+  private getDetailModel(record: unknown): object {
+    const detailModel = this.api.createChild<object>(this.detailTemplate, this);
+    this.api.setVariable('record', record, detailModel);
```
aaron-steinfeld, line 44: how can we avoid this being some magic variable that needs to be known? One option would be to have the variable name actually be set as an input. Alternatively if we do keep it, I think the name should be super generic like `source` or `input`. Hardcoding it though would, apart from discoverability, also force bad things like reserved variable names and an inability to nest. Those aren't the end of the world, but easily avoidable.

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.ts
```diff
@@ -0,0 +1,31 @@
+import { PortalInjector } from '@angular/cdk/portal';
+import { Injectable, Injector } from '@angular/core';
+import { OverlayService, PopoverRef, SheetSize } from '@hypertrace/components';
+import {
+  DetailSheetInteractionContainerComponent,
+  DETAIL_SHEET_INTERACTION_MODEL
+} from './container/detail-sheet-interaction-container.component';
+
+@Injectable({
+  providedIn: 'root'
```
aaron-steinfeld, line 10: one of the dependencies of this, overlay service, is not a root injectable (because it requires an injector). So this probably shouldn't be either.

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.ts
```diff
@@ -0,0 +1,31 @@
...
+      {
+        content: DetailSheetInteractionContainerComponent,
+        size: sheetSize
+      },
+      this.buildDetailOverlayInjector(detailModel)
+    );
+  }
+
+  private buildDetailOverlayInjector(detailModel: object): Injector {
+    return new PortalInjector(
```
aaron-steinfeld, line 26: Up to you, but I've actually moved away from using portal injector when not directly work with CDK - newer angular APIs make it easy enough to use injector directly -
```
    return Injector.create({
      providers: [
        {
          provide: DETAIL_SHEET_INTERACTION_MODEL,
          useValue: detailModel
        }
      ],
      parent: this.injector
    });
```

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
anandtiwary: Yes. I had exactly the former in mind. I will make the variable name as an input.

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.ts
anandtiwary: Yes. I agree. Wanted to confirm this in PR review. I would provide the service at the module level.

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.ts
anandtiwary: Sure. Will use this

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
anandtiwary: yep

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
anandtiwary: Yeah. We may add more features to this container which would be configurable through the interaction model.

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/container/detail-sheet-interaction-container.component.ts
anandtiwary: okay

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
anandtiwary: So, I kept the name detail because of its ability to provide the original data which was interacted with(row in case of table). So just with just overlay sheet it is not obv. I am happy to rename this, but i want it to convey that this model injects a context data.

anandtiwary:
Tests incoming

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction.module.ts
```diff
@@ -2,13 +2,15 @@ import { NgModule } from '@angular/core';
...
 
 @NgModule({
   imports: [
     DashboardCoreModule.with({
       models: [DetailSheetInteractionHandlerModel]
     })
   ],
-  declarations: [DetailSheetInteractionContainerComponent]
+  declarations: [DetailSheetInteractionContainerComponent],
+  providers: [DetailSheetInteractionHandlerService]
```
aaron-steinfeld, line 14: It won't make a difference here, but I was reading somewhere about avoiding declaring providers this way. Instead, if you declare it in the other direction (i.e.  `@Injectable({providedIn: DetailSheetInteractionModule})`) then it allows it to be tree shaken if not used. We know this one is used, but good habit to get in assuming it works.

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  test('should call overlay service', () => {
+    spectator.service.showSheet({}, SheetSize.Medium);
+    expect(spectator.inject(OverlayService).createSheet).toHaveBeenCalledWith(
+      {
+        content: DetailSheetInteractionContainerComponent,
+        size: SheetSize.Medium
+      },
+      expect.objectContaining({
+        /* Injector */
```
aaron-steinfeld, line 30:
```
      expect.any(Injector)
```

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
```diff
@@ -18,30 +18,36 @@ export class DetailSheetInteractionHandlerModel implements InteractionHandler {
...
-  public execute(record: unknown): Observable<void> {
-    this.handlerService.showSheet(this.getDetailModel(record), this.sheetSize);
+  public execute(source: unknown): Observable<void> {
+    this.handlerService.showSheet(this.getDetailModel(source), this.sheetSize);
 
     return of();
   }
 
-  private getDetailModel(record: unknown): object {
+  private getDetailModel(source: unknown): object {
```
aaron-steinfeld, line 48: createChild will default to this model as the parent

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction.module.ts
anandtiwary: I don't think it will ever get tree shaken. We include the model in the common tracing dashboard interactions module which is also imported in the dashboard modules. Also, I would prefer it to be this way till we make a sweeping changes across all similar services

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.test.ts
anandtiwary: done

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
anandtiwary: removed

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction.module.ts
aaron-steinfeld: OK - although in general we should separate the idea of "will it be tree shaken" from "can it be shaken". We want to support the latter whenever possible since the former may change.

aaron-steinfeld:
I take it you're not running these locally? 😛

anandtiwary:
No.. didn't want to run npm install again. but had to do it now. The ` expect.any(Injector)` doesn't seem to work

anandtiwary:
```
- Any<Injector>,
    + {"_destroyed": false, "injectorDefTypes": Set {}, "onDestroy": Set {}, "parent": {"_destroyed": false, "injectorDefTypes": [Set], "onDestroy": [Set], "parent": [R3Injector], "records": [Map], "scope": "root", "source": "DynamicTestModule"}, "records": Map {[InjectionToken] => [Object], [InjectionToken] => [Object]}, "scope": null, "source": null},

    Number of calls: 1
```

anandtiwary:
I think it has nested objects and due to that the expect matcher api doesn't work. I had to move to that empty object to pass the test. There were no useful instance property too that I could use.

anandtiwary:
okay. have a better test now.

aaron-steinfeld:
> I think it has nested objects and due to that the expect matcher api doesn't work. I had to move to that empty object to pass the test. There were no useful instance property too that I could use.

Strange - expected does work nested, but the issue is that apparently angular's ivy injector is not actually a subclass of injector, they must just do a switcheroo at runtime. 
From core types:
```typescript
declare class R3Injector {
    readonly parent: Injector;
...
}
```

PR #77, opened 2020-08-13
feat: always showing header and adding optional title property

anandtiwary:


projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.model.ts
```diff
@@ -27,6 +27,13 @@ export class DetailSheetInteractionHandlerModel implements InteractionHandler {
   })
   public sheetSize: SheetSize = SheetSize.Large;
 
+  @ModelProperty({
+    key: 'title-property-path',
+    required: false,
+    type: STRING_PROPERTY.type
+  })
+  public titlePropertyPath?: string;
```
anandtiwary, line 35: Trying to make use of existing sheet overlay default header as all sheets should show a close button. Since the sheet title can be derived from the data, I thought of providing this path property. This would make it easier to extract title from data itself. 

If we don't do this, then the content widget itself has to provided title and close button, which i think would be pain.

anandtiwary: Please note, we are using the `source` data for this. This may not work in cases where we want to show a title which is based on the new data fetched by the content widget.

anandtiwary: Screen Shot 2020-08-12 at 5 42 33 PM

aaron-steinfeld: Implementation is fine, but this feels a bit weird to me. I wonder if we can leverage variables with this too? i.e. just provide a title input on the child model, which can be populated via `title: "${source.titlePath}"`. But if the child is opaque and the title needs to go into the sheet service, may not be feasible. Anyway, food for thought, not a blocker.

anandtiwary: Yeah. title needs to go in sheet service so had to choose this option. Anyway, I can revisit if I get a better idea later.

PR #79, opened 2020-08-13
fix: more changes to support overlay with table

anandtiwary:
- Overlay service should be root injectable. It should track the currently active
  sheet/modal popover and ensure that only one sheet/modal instance exist at a
  given time. This helps in avoiding duplicate popover on the screen.
  In general, Only one sheet or modal should be visible at a given time anyway.

- Added a new size to sheet

- Single mode table selection should clear out any previous selection

- Multi selection should store the updated row selection post toggle

### DIFF
--- a/projects/components/src/overlay/overlay.service.ts
+++ b/projects/components/src/overlay/overlay.service.ts
@@ -8,8 +8,12 @@ import { ModalOverlayConfig } from './modal/modal';
 import { ModalOverlayComponent } from './modal/modal-overlay.component';
 import { SheetOverlayComponent } from './sheet/sheet-overlay.component';
 
-@Injectable()
+@Injectable({
+  providedIn: 'root'
+})
 export class OverlayService {
+  private activePopover?: PopoverRef;
+
   public constructor(private readonly popoverService: PopoverService, private readonly defaultInjector: Injector) {}
 
   public createSheet(config: SheetOverlayConfig, injector: Injector = this.defaultInjector): PopoverRef {
@@ -23,10 +27,16 @@ export class OverlayService {
       data: config
     });
 
+    popover.closeOnNavigation();
+
+    this.setActivePopover(popover);
+
     return popover;
   }
 
   public createModal(config: ModalOverlayConfig, injector: Injector = this.defaultInjector): PopoverRef {
+    this.activePopover?.close();
+
     const popover = this.popoverService.drawPopover({
       componentOrTemplate: ModalOverlayComponent,
       parentInjector: injector,
@@ -37,6 +47,15 @@ export class OverlayService {
       data: config
     });
 
+    popover.closeOnNavigation();
+
+    this.setActivePopover(popover);
+
     return popover;
+  }
+
+  private setActivePopover(popover: PopoverRef): void {
+    this.activePopover?.close();
+    this.activePopover = popover;
   }
 }
--- a/projects/components/src/overlay/overlay.service.ts
+++ b/projects/components/src/overlay/overlay.service.ts
@@ -1,9 +1,10 @@
-import { Injectable, Injector } from '@angular/core';
+import { Injectable, Injector, OnDestroy } from '@angular/core';
 import { PopoverFixedPositionLocation, PopoverPositionType } from '../popover/popover';
 import { PopoverRef } from '../popover/popover-ref';
 import { PopoverService } from '../popover/popover.service';
 import { SheetOverlayConfig } from './sheet/sheet';
 
+import { Subscription } from 'rxjs';
 import { ModalOverlayConfig } from './modal/modal';
 import { ModalOverlayComponent } from './modal/modal-overlay.component';
 import { SheetOverlayComponent } from './sheet/sheet-overlay.component';
@@ -11,10 +12,15 @@ import { SheetOverlayComponent } from './sheet/sheet-overlay.component';
 @Injectable({
   providedIn: 'root'
 })
-export class OverlayService {
+export class OverlayService implements OnDestroy {
   private activePopover?: PopoverRef;
+  private closeSubscription?: Subscription;
 
   public constructor(private readonly popoverService: PopoverService, private readonly defaultInjector: Injector) {}
+
+  public ngOnDestroy(): void {
+    this.closeSubscription?.unsubscribe();
+  }
 
   public createSheet(config: SheetOverlayConfig, injector: Injector = this.defaultInjector): PopoverRef {
     const popover = this.popoverService.drawPopover({
@@ -55,7 +61,11 @@ export class OverlayService {
   }
 
   private setActivePopover(popover: PopoverRef): void {
+    this.closeSubscription?.unsubscribe();
     this.activePopover?.close();
+
     this.activePopover = popover;
+    this.activePopover.closeOnNavigation();
+    this.closeSubscription = this.activePopover.closed$.subscribe(() => (this.activePopover = undefined));
   }
 }
--- a/projects/components/src/overlay/sheet/sheet-overlay.component.scss
+++ b/projects/components/src/overlay/sheet/sheet-overlay.component.scss
@@ -20,11 +20,17 @@
   &.sheet-size-small {
     width: 320px;
   }
+
   &.sheet-size-medium {
     width: 600px;
   }
+
   &.sheet-size-large {
     width: 840px;
+  }
+
+  &.sheet-size-extra-large {
+    width: 1280px;
   }
 
   .header {
--- a/projects/components/src/table/table.component.ts
+++ b/projects/components/src/table/table.component.ts
@@ -175,7 +175,7 @@ export class TableComponent
     width: '32px',
     visible: true,
     renderer: StandardTableCellRendererType.Checkbox,
-    onClick: (row: StatefulTableRow) => this.toggleRowSelected(row)
+    onClick: (row: StatefulTableRow) => this.toggleRowSelectedForMulti(row)
   };
 
   public readonly expandedDetailColumnConfig: TableColumnConfig = {
@@ -229,6 +229,9 @@ export class TableComponent
 
   @Output()
   public readonly selectionsChange: EventEmitter<StatefulTableRow[]> = new EventEmitter<StatefulTableRow[]>();
+
+  @Output()
+  public readonly rowClicked: EventEmitter<StatefulTableRow> = new EventEmitter<StatefulTableRow>();
 
   @Output()
   public readonly hoveredChange: EventEmitter<StatefulTableRow | undefined> = new EventEmitter<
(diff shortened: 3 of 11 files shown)


projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
```diff
@@ -59,4 +60,8 @@ export class TableWidgetRendererComponent
   public get syncWithUrl(): boolean {
     return this.model.style === TableStyle.FullPage;
   }
+
+  public onRowSelection(selections: TableRow[]): void {
+    this.api.model.selectionHandler?.execute(selections[0]);
```
anandtiwary, line 65: This is weird. Since the output is same for both single and multi selection, at this point, how should we handle the action? Above change would work for single selection, but if someone enables multi selection on the table and there are multiple row selections, should we just ignore such a case? Also, selections[0] will not always be the latest selected row.

anandtiwary: So considering that when we click on a selected row it gets unselected, what should be the behavior of the sheet overlay? Should it also gets closed?  Should it stay open while the row selection style is removed?

projects/components/src/overlay/overlay.service.ts
```diff
@@ -8,8 +8,12 @@ import { ModalOverlayConfig } from './modal/modal';
 import { ModalOverlayComponent } from './modal/modal-overlay.component';
 import { SheetOverlayComponent } from './sheet/sheet-overlay.component';
 
-@Injectable()
+@Injectable({
+  providedIn: 'root'
```
aaron-steinfeld, line 12: If providing this in root, the default injector should be removed - it needs to be provided.

projects/components/src/overlay/overlay.service.ts
```diff
@@ -37,6 +47,15 @@ export class OverlayService {
...
+    popover.closeOnNavigation();
+
+    this.setActivePopover(popover);
+
     return popover;
   }
+
+  private setActivePopover(popover: PopoverRef): void {
+    this.activePopover?.close();
+    this.activePopover = popover;
```
aaron-steinfeld, line 59: this should be unset on popover close

projects/components/src/overlay/sheet/sheet-overlay.component.scss
```diff
@@ -20,13 +20,19 @@
...
+
   &.sheet-size-medium {
     width: 600px;
   }
+
   &.sheet-size-large {
     width: 840px;
   }
 
+  &.sheet-size-extra-large {
```
aaron-steinfeld, line 32: Isn't this bigger than the widget we even design for?

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.ts
```diff
@@ -5,7 +5,9 @@ import {
   DETAIL_SHEET_INTERACTION_MODEL
 } from './container/detail-sheet-interaction-container.component';
 
-@Injectable()
+@Injectable({
+  providedIn: 'root'
```
aaron-steinfeld, line 9: likewise, shouldn't be in root if it's providing its own injector

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
aaron-steinfeld: I'd think we'd just give the whole array to the selection handler - it's on the handler to deal with it.
As to close behavior, I'd say it should close either on any external click, or none - wouldn't special case the deselect case.

projects/components/src/table/table.component.ts
```diff
@@ -421,10 +421,18 @@ export class TableComponent
     this.changeDetector.markForCheck();
   }
 
-  public toggleRowSelected(row: StatefulTableRow): void {
+  public toggleExistingOrSelectNewRow(row: StatefulTableRow): void {
+    const rowIndexInSelections = (this.selections ?? []).findIndex(selection => isEqualIgnoreFunctions(selection, row));
+    rowIndexInSelections >= 0 ? (this.selections = []) : (this.selections = [row]);
```
aaron-steinfeld, line 426: It's not clear to me how this is different than the multi case? Even if they are slightly different, why do we need two functions?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
anandtiwary: I have decided to hide it as the row is un selected. Lemme know if you disagree

projects/components/src/overlay/overlay.service.ts
anandtiwary: Yeah. Is this really a problem though?  I suppose I would have to subscribe to popoverRef.closed$ which I want to avoid.

projects/components/src/overlay/sheet/sheet-overlay.component.scss
anandtiwary: Hah.. Few things are in flux. I spoke to Michael about it. Yes, It looks too big for laptop screen. 

There would be some more changes around this as he needs more time to play with it. Happy to revert this later.

projects/components/src/table/table.component.ts
anandtiwary: In Multi, if you click on second row then selections array will have two rows. In Single selection, if you click on a new row, the previous row should be unselected or toggled and only the current row should now be selected. Created another method to keep logic separate.

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.ts
anandtiwary: I have to understand this. You can explain above

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
anandtiwary: @aaron-steinfeld that makes sense. Since I am using detail sheet handler, I can't put any selection specific code in the handler. 
However, I can put a restriction that if it gets empty or undefined data, then it would clear out its state/dispose any popover etc. I can see it work in that case. Do you think its a good assumption?

projects/components/src/overlay/overlay.service.ts
aaron-steinfeld: Looked into this a bit, I think it's OK to default it after all. My understanding _was_ that because these are used to resolve a component factory, only components available in the root module could be instantiated. It looks like that's not the case (maybe it changed with ivy, maybe it never was). I suspect it does still hold with providers, but we use view/local providers less frequently, so we can get away with defaulting I think.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
aaron-steinfeld: Yes more or less - I'd have it take an array of values, and then the specific handler can choose how to react to that. This handler can close any popups if receiving an empty array, open/replace if receiving a single, and error if receiving multiple.

projects/components/src/overlay/overlay.service.ts
aaron-steinfeld: Always hesitant to use subscriptions, but since this is long lived and the subscription is short (and we're holding the popover state anyway), I think subscribing there is just fine. Just becomes a trivial callback of `popover.closed$.subscribe(() => this.clear())` or something

projects/distributed-tracing/src/shared/dashboard/interaction/detail-sheet/detail-sheet-interaction-handler.service.ts
aaron-steinfeld: took it back above.

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-renderer.component.ts
anandtiwary: nice. This works.

projects/components/src/overlay/overlay.service.ts
anandtiwary: okay. Thanks for explaining!

projects/components/src/overlay/overlay.service.ts
anandtiwary: not that simple. It should only clear if popover was closed from outside. On line 58, the close call should not reset this.

projects/components/src/table/table.component.ts
aaron-steinfeld: Got it - but rather than fork the functions externally, I think it would be much clearer to keep the logic internal to the call so callers don't have to make the determination of the selection mode before deciding which to call. I killed the mutation too because I'm a freak, but I find this much more readable. Probably because everyone finds their own code more readable, so up to you.
```typescript
  public toggleRowSelected(toggledRow: StatefulTableRow): void {
    const previousSelections = this.selections ?? [];
    const deselectedRow = previousSelections.find(selection => isEqualIgnoreFunctions(selection, toggledRow));
    if (deselectedRow) {
      this.selections = without(previousSelections, deselectedRow);
    } else {
      this.selections = this.hasMultiSelect() ? [...previousSelections, toggledRow] : [toggledRow];
    }
    this.selectionsChange.emit(this.selections);
    this.changeDetector.markForCheck();
  }
```

projects/components/src/overlay/overlay.service.ts
anandtiwary: I am thinking this.
```
  private setActivePopover(popover: PopoverRef): void {
    this.closeSubscription?.unsubscribe();
    this.activePopover?.close();

    this.activePopover = popover;
    this.activePopover.closeOnNavigation();
    this.closeSubscription = this.activePopover.closed$.subscribe(() => this.activePopover = undefined);
  }
```

aaron-steinfeld: why not? line 58 is the case where a new popover is opened without closing the previous one, right? So 58 would be called, the result of which would clear active popover via that subscription. 59 would then execute and set the now undefined value. This could be problematic if closed$ were async, but it's not. If we want to be super careful, we could also just make it to be something like `if (this.activePopover === closedPopover) { this.activePopover = undefined;}`

aaron-steinfeld: Your suggestion works too, up to you (sorry github really needs to notify me of new comments)

projects/components/src/table/table.component.ts
anandtiwary: Yeah. This is much more cleaner. I got lazy when writing the previous code.

anandtiwary: done

PR #85, opened 2020-08-18
feat: greeting label component and widget

arjunlalb:
Defining greeting label component and widget renderer.

Screenshot 2020-08-18 at 1 48 27 PM


projects/components/src/greeting-label/greeting-label.component.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { GreetingLabelComponent } from '@hypertrace/components';
+import { createHostFactory, Spectator } from '@ngneat/spectator/jest';
+
+describe('Greeting Label component', () => {
```
aaron-steinfeld, line 4: file name: `...test.ts`

projects/components/src/greeting-label/greeting-label.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  public greetingMessage: string = '';
+
+  public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (changes.label) {
+      this.setGreetingMessage();
+    }
+  }
+
+  private setGreetingMessage(): void {
+    const parts = this.label.split('{greeting}');
```
aaron-steinfeld, line 23: I don't love this - it means the caller needs to know the implementation to know how to follow the specific template rules of this component. Do we ever envision this being used where the greeting comes anywhere but the beginning of the message? If not, then maybe we just remove the templating from the label message itself. Because I could also argue that I could use `{greeting}` twice in my message.

projects/dashboards/src/widgets/greeting-label/greeting-label-widget-renderer.component.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { GreetingLabelComponent } from '@hypertrace/components';
+import { RENDERER_API } from '@hypertrace/hyperdash-angular';
+import { createComponentFactory, Spectator } from '@ngneat/spectator/jest';
+import { EMPTY } from 'rxjs';
+import { GreetingLabelWidgetRendererComponent } from './greeting-label-widget-renderer.component';
+import { GreetingLabelWidgetModel } from './greeting-label-widget.model';
+
+describe('Highlighted label widget renderer component', () => {
```
aaron-steinfeld, line 8: describe is from a different component. also file name.

projects/components/src/greeting-label/greeting-label.component.spec.ts
arjunlalb: oops. thats ng generate. will fix.

projects/components/src/greeting-label/greeting-label.component.ts
arjunlalb: As of now, we don't have a use case for it to be anywhere other than the beginning. But would the flexibility hurt? In case we decide to use it at other locations in the future, this could be handy.
As for the possibility of using `{greeting}` twice in the message, we can obviously add a check against it, but that'd be an overkill IMO. I think its ok to leave that upto the user to use it correctly.

aaron-steinfeld: IMO both points go together.
> we don't have a use case for it to be anywhere other than the beginning. But would the flexibility hurt?

We're providing flexibility for one scenario but not another, and neither are needed. Generally, we want the API to be as simple as we can to support the use cases we have. When adding more functionality/flexibility, then it should be clear and consistent (not going to lie and try to argue we're always the best about this). Here, I feel like it makes the API more confusing (because now as a caller you need to understand how to formulate a template string - does the widget include the comma or do I?! what's the template var name? Should I capitalize?) and more non-standard/unexpected (because unlike a standard template string, you can only reference the template var once)

projects/dashboards/src/widgets/greeting-label/greeting-label-widget-renderer.component.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    ]
+  });
+
+  beforeEach(() => {
+    mockModel.labelTemplate = '{greeting}, test';
+    spectator = createComponent();
+  });
+
+  test('should render greeting label correctly', () => {
+    spyOn(Date.prototype, 'getHours').and.returnValue(9);
```
aaron-steinfeld, line 35: I'm confused on this test - shouldn't 9 be morning? I'm assuming this is because the greeting message is already set before we spied here (it'd be set on 31), so the spy has no effect. But I don't think that's the intention here, and if it is A. the test isn't clear on it and B. that would make this test produce different results at different times of day which is of course, not good.

arjunlalb: That was a mistake. Fixed it.

projects/components/src/greeting-label/greeting-label.component.ts
arjunlalb: Assuming we don't want the greeting to be templatized, the problem about understanding the API still persists.
example: if I want to show `Good morning, here is your report`, assuming that greeting is always in the beginning, the consumer still needs to know if the user supplied template has to be with or without the comma.
So, not templatizing the label doesn't solve that problem.

For now, I'll make this simple and go with prefixing the greeting.

aaron-steinfeld: > assuming that greeting is always in the beginning, the consumer still needs to know if the user supplied template has to be with or without the comma.
So, not templatizing the label doesn't solve that problem.

Abstractly, I think of it as just a message that is provided. The widget could choose to use a common, a period and new line, whatever - the widget's concern is to put the provided message in a greeting context. So I guess I'd think the comma belongs to the greeting too, and the message can be just `here is your report` (the greeting widget handling capitalization as appropriate). Anyway, this is such a strange and special cased widget, that we probably don't have to worry too much about this.

projects/components/src/greeting-label/greeting-label.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  private setGreetingMessage(): void {
+    this.greetingMessage = `${this.getGreeting()}${this.suffixLabel}`;
+  }
+
+  private getGreeting(): string {
+    const hour = new Date().getHours();
+
+    if (hour < 12) {
+      return 'Good Morning';
```
aaron-steinfeld, line 30: Some nitpicky suggestions at the time windows:
Morning 4AM-12PM Afternoon 12PM-5PM Evening otherwise (5PM-4AM)

arjunlalb: I have to send another PR to include this widget in home dash. Will make these changes along with that.

PR #87, opened 2020-08-19
fix: fixing detail overlay issues

anandtiwary:
- injecting Layout change service in popover
- adding a model property to hide header


projects/assets-library/assets/styles/_interaction.scss
```diff
@@ -10,6 +10,7 @@
 @mixin card-hover {
   &:hover {
     @include dropdown-box-shadow();
+    cursor: pointer;
```
jake-bassett, line 13: Wouldn't we want the cursor to always be there? Not just on hover?

I understand that by definition it would never show unless we were hovering the element, but I'm coming from a standpoint of correctness.

anandtiwary: Yes. That's true, but if we don't do it here it is easy to miss that the card hover doesn't bring the cursor style.

PR #88, opened 2020-08-20
feat: add support for IN operator in filter bar

jake-bassett:


projects/components/src/filter-bar/filter/parser/filter-parser.service.ts
```diff
@@ -24,7 +24,8 @@ export class FilterParserService {
   public constructor(private readonly filterBuilderService: FilterBuilderService) {}
 
   public parseUserFilterString(filterString: string, attribute: FilterAttribute): Filter | undefined {
-    const parsed = this.parseFilterString(filterString, attribute, USER_FILTER_OPERATORS, attribute.displayName);
+    const spacedOperators = USER_FILTER_OPERATORS.map(op => ` ${op} `);
```
aaron-steinfeld, line 27: why is the space needed?

projects/components/src/filter-bar/filter/builder/string-filter-builder.ts
```diff
@@ -2,22 +2,42 @@ import { FilterType } from '../../filter-type';
 import { UserFilterOperator } from '../filter-api';
 import { AbstractFilterBuilder } from './abstract-filter-builder';
 
-export class StringFilterBuilder extends AbstractFilterBuilder<string | undefined> {
-  public convertValue(value: unknown): string | undefined {
-    return value === undefined ? undefined : String(value);
+export class StringFilterBuilder extends AbstractFilterBuilder<string | string[] | undefined> {
```
aaron-steinfeld, line 5: Should we be adding support to numeric filter too? number attributes should support in right?

projects/components/src/filter-bar/filter/parser/filter-parser.service.ts
jake-bassett: Since we intend to split on the operator. In this case is "in", which is also often in values and keys.

The filter pills enforce a space around the operator so this will work for now, but I'm not a fan of relying on the component for validation. This feels too brittle. We really need a full on parser or there are going to be corner cases we miss.

projects/components/src/filter-bar/filter/builder/string-filter-builder.ts
jake-bassett: IN could make sense for numbers, especially since things like status code are numbers. I will add them in a follow up, but wanted to unblock myself for column filtering first.

projects/components/src/filter-bar/filter/parser/filter-parser.service.ts
aaron-steinfeld: If only we had a parsing library.. :)

jake-bassett: I still need to research more, but from what I saw when I started investigating, we need a little more than parjs as it missed the same use cases my own string parsing did.

PR #89, opened 2020-08-20
feat: remove predefined time greater than 1 month

jake-bassett:


anandtiwary:
Great! I was just asking someone about this. What about custom time range? Should that have any bounds?

jake-bassett:
> Great! I was just asking someone about this. What about custom time range? Should that have any bounds?

I don't see any reason to put bounds on custom time range at the moment, but worth keeping an eye out if it becomes an issue.

PR #90, opened 2020-08-21
fix: fix page scrolling in safari

arjunlalb:


projects/assets-library/assets/styles/_layout.scss
```diff
@@ -12,6 +12,7 @@
   min-height: 0;
   padding: $padding-top $padding-sides $padding-bottom;
   overflow: auto;
+  max-height: -webkit-fill-available; // For safari
```
skjindal93, line 15: https://github.com/hypertrace/hypertrace-core-ui/pull/25#discussion_r458700095

arjunlalb: I remember the comment. But couldn't find an alternate way around this.

PR #100, opened 2020-08-24
fix: conditionally apply fixed/sticky position for pagination

arjunlalb:


projects/components/src/table/table.component.ts
```diff
@@ -141,7 +141,11 @@ import {
     </div>
 
     <!-- Pagination -->
-    <div class="pagination-controls" *ngIf="this.pageable">
+    <div
+      class="pagination-controls"
+      *ngIf="this.pageable"
+      [style.position]="this.display === 'full-page' ? 'fixed' : 'sticky'"
```
skjindal93, line 147: Would prefer to use the enum for full page

arjunlalb: @skjindal93 done.

projects/components/src/table/table.component.ts
```diff
@@ -55,7 +55,7 @@ import {
     <cdk-table
       [multiTemplateDataRows]="this.isDetailType()"
       [dataSource]="this.dataSource"
-      [ngClass]="[this.display, this.pageable ? 'bottom-margin' : '']"
+      [ngClass]="[this.display, this.pageable && this.display === '${TableStyle.FullPage}' ? 'bottom-margin' : '']"
```
skjindal93, line 58: probably define a boolean flag the component level and then it can be used at both places? `isTableFullPage`

arjunlalb: @skjindal93 done

projects/components/src/table/table.component.ts
```diff
@@ -273,13 +273,17 @@ export class TableComponent
...
+  public isTableFullPage: boolean;
+
   public dataSource?: TableCdkDataSource;
 
   public constructor(
     private readonly changeDetector: ChangeDetectorRef,
     private readonly navigationService: NavigationService,
     private readonly activatedRoute: ActivatedRoute
   ) {
+    this.isTableFullPage = this.display === TableStyle.FullPage;
```
aaron-steinfeld, line 285: this check should be in ngOnChanges since it could technically change (although I admit, unlikely to)

arjunlalb: I omitted it there because TableStyle won't change across digest cycles. Will add it there.

projects/components/src/table/table.component.ts
```diff
@@ -293,6 +293,10 @@ export class TableComponent
   }
 
   public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (changes.display) {
+      this.isTableFullPage = this.display === TableStyle.FullPage;
```
aaron-steinfeld, line 297: This should replace line 285 - this.display will never be set in the constructor (inputs come in later)

skjindal93 (approved):

PR #107, opened 2020-08-25
feat: add support for styling widgets

arjunlalb:
Screenshot 2020-08-25 at 5 45 21 PM

### DIFF
--- a/projects/dashboards/src/widgets/container/layout/auto/auto-container-layout.model.ts
+++ b/projects/dashboards/src/widgets/container/layout/auto/auto-container-layout.model.ts
@@ -41,6 +41,7 @@ export class AutoContainerLayoutModel extends ContainerLayout {
       columns: `repeat(${gridDimension[1]}, minmax(${this.minColumnWidth}px, 1fr))`,
       gap: `${this.gridGap}`,
       enableStyle: this.enableStyle,
+      widgetTheme: this.widgetTheme,
       children: children.map(child => ({ model: child, areaSpan: '' }))
     };
   }
--- a/projects/dashboards/src/widgets/container/layout/widget-theme.ts
+++ b/projects/dashboards/src/widgets/container/layout/widget-theme.ts
@@ -1,0 +1,4 @@
+export enum WidgetTheme {
+  Dark = 'dark',
+  Light = 'light'
+}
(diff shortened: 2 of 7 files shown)


projects/dashboards/src/widgets/container/layout/widget-theme.ts
```diff
@@ -0,0 +1,4 @@
+export enum WidgetTheme {
+  Dark = 'dark',
```
skjindal93, line 2: Should we call it Gray, instead of Dark? Dark theme gives me a Dark mode feel

arjunlalb: Done.

projects/dashboards/src/widgets/container/layout/auto/auto-container-layout.model.ts
```diff
@@ -41,6 +41,7 @@ export class AutoContainerLayoutModel extends ContainerLayout {
       columns: `repeat(${gridDimension[1]}, minmax(${this.minColumnWidth}px, 1fr))`,
       gap: `${this.gridGap}`,
       enableStyle: this.enableStyle,
+      widgetTheme: this.widgetTheme,
```
aaron-steinfeld, line 44: I'm super hesitant to add theme to our existing model considering it's already built in (but not fleshed out). @arjunlalb I'm guessing you were unaware of it, so do you mind looking into that first and seeing if we could accomplish the same thing with that system?

Some pointers:
- It's a sibling to the data source system, so it can be accessed in a similar way - `this.api.getTheme()` either in the model or renderer apis.
- A theme value can be accessed directly in the DOM through a pipe (we should probably rename this) - https://github.com/hypertrace/hyperdash-angular/blob/main/projects/hyperdash-angular/src/rendering/theme-property.pipe.ts
- It's defined on a model in the same way as data, as a special prop on any model that is inherited. A working example of this can be found in the ancient hyperdash playground: https://github.com/hypertrace/hyperdash-angular/blob/main/src/example-dash/example-dash.component.ts . We can subclass and create new theme models beyond the built in one which was more of a POC.

arjunlalb: aah. You're right. I wasn't aware of it. Will take a look.

Is that a theme system for the whole app or usable for just isolated usecases?

aaron-steinfeld: We can use it isolated but it's meant to integrate with the broader app. Specifically two aspects to watch out for:
- Themes are inherited. If you theme a container, then any of its children will inherit that theme by default. They're also merged.
- Using a theme is opt in for the renderer. Because different widgets will apply styles differently, they need to explicitly apply theme values.

PR #109, opened 2020-08-25
feat: in filter button

jake-bassett:


projects/components/src/filter-button/filter-button.service.ts
```diff
@@ -19,10 +19,18 @@ export class FilterButtonService {
     return this.filterBuilderService.lookup(attribute).buildFiltersForAvailableOperators(attribute, value);
   }
 
+  public isSupportedOperator(attribute: FilterAttribute, operator: UserFilterOperator): boolean {
+    return this.filterBuilderService.lookup(attribute).supportedOperators().includes(operator);
+  }
+
   /*
    * TODO: Combine service with filter-bar.service. A few duplicated methods here.
```
jake-bassett, line 27: PR to refactor file structure and combine service methods (comment on line 27) to follow but wanted to keep this PR scope down.

projects/components/src/checkbox/checkbox.component.ts
```diff
@@ -13,7 +13,7 @@ import { MatCheckboxChange } from '@angular/material/checkbox';
       [disabled]="this.disabled"
       class="htc-checkbox"
     >
-      <htc-label class="label" *ngIf="this.label" [label]="this.label"></htc-label>
+      <htc-label class="label" *ngIf="this.label !== undefined" [label]="this.label"></htc-label>
```
anandtiwary, line 16: Why switch to only undefined? It can be empty as well.

projects/components/src/filter-button/in-filter-button.component.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  public isSupported: boolean = false;
+  public selected: Set<T> = new Set<T>();
+
+  public constructor(private readonly filterButtonService: FilterButtonService) {}
+
+  public ngOnChanges(changes: TypedSimpleChanges<this>): void {
+    if (changes.attribute) {
+      this.isSupported = this.attribute
+        ? this.filterButtonService.isSupportedOperator(this.attribute, UserFilterOperator.In)
```
anandtiwary, line 57: I wish there was a way to remove these codecov messages. :P 
Makes it hard to read

projects/components/src/filter-button/in-filter-button.component.ts
```diff
@@ -0,0 +1,97 @@
...
+import { IconSize } from '../icon/icon-size';
+import { FilterButtonService } from './filter-button.service';
+
+@Component({
+  selector: 'htc-in-filter-button',
+  styleUrls: ['./in-filter-button.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div class="filter-button" *ngIf="this.isSupported">
+      <htc-popover [closeOnClick]="false" (popoverOpen)="this.onPopoverOpen()" (popoverClose)="this.onPopoverClose()">
```
arjunlalb, line 15: can we use the multi-select component here?

projects/components/src/checkbox/checkbox.component.ts
jake-bassett: The previous check was not working for 0, so refined it. I can add a check for empty strings as well.

projects/components/src/filter-button/in-filter-button.component.ts
jake-bassett: Agreed!

projects/components/src/filter-button/in-filter-button.component.ts
jake-bassett: I can give it a shot and see.

jake-bassett: Ah, just looked at what that component is. The multiselect dropdown isn't quite what we want here. Just simple checkboxes.

projects/components/src/table/renderers/header-cell/table-header-cell-renderer.component.ts
```diff
@@ -16,11 +17,28 @@ import { TableCellRendererService } from '../table-cell-renderer.service';
...
-      {{ this.columnConfig.title }}
+      <htc-in-filter-button
+        class="filter-button"
+        *ngIf="this.columnConfig?.filterAttribute && this.leftAlignFilterButton"
+        [metadata]="this.metadata"
+        [attribute]="this.columnConfig.filterAttribute"
+        [values]="this.values"
+      ></htc-in-filter-button>
+      <div class="title" [ngClass]="this.classes" (click)="this.sortChange.emit()">{{ this.columnConfig.title }}</div>
+      <htc-in-filter-button
```
aaron-steinfeld, line 28: Template!

Example from the data cell change since I asked there too:
```diff
diff --git a/projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts b/projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
index 56b3ec7..e6dd6c2 100644
--- a/projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
+++ b/projects/components/src/table/renderers/data-cell/table-data-cell-renderer.component.ts
@@ -21,26 +21,25 @@ import { TableCellRendererService } from '../table-cell-renderer.service';
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <div class="table-data-cell-renderer" [class.selected]="this.popoverOpen">
-      <htc-filter-button
-        class="filter-button"
-        *ngIf="this.columnConfig?.filterAttribute && this.leftAlignFilterButton"
-        [metadata]="this.metadata"
-        [attribute]="this.columnConfig?.filterAttribute"
-        [value]="this.parseValue()"
-        (popoverOpen)="this.popoverOpen = $event"
-      ></htc-filter-button>
+      <ng-container *ngIf="this.columnConfig?.filterAttribute && this.leftAlignFilterButton">
+        <ng-container *ngTemplateOutlet="filterButton"></ng-container>
+      </ng-container>
       <div class="cell-renderer-content" [ngClass]="this.alignment" (click)="this.onClick()">
         <ng-container #cellRenderer></ng-container>
       </div>
+      <ng-container *ngIf="this.columnConfig?.filterAttribute && !this.leftAlignFilterButton">
+        <ng-container *ngTemplateOutlet="filterButton"></ng-container>
+      </ng-container>
+    </div>
+    <ng-template #filterButton>
       <htc-filter-button
         class="filter-button"
-        *ngIf="this.columnConfig?.filterAttribute && !this.leftAlignFilterButton"
         [metadata]="this.metadata"
         [attribute]="this.columnConfig?.filterAttribute"
         [value]="this.parseValue()"
         (popoverOpen)="this.popoverOpen = $event"
       ></htc-filter-button>
-    </div>
+    </ng-template>
   `
 })
 export class TableDataCellRendererComponent implements OnInit {

```

projects/components/src/filter-button/filter-button.service.ts
```diff
@@ -37,17 +45,29 @@ export class FilterButtonService {
     this.setUrlFilters([...filters]);
   }
 
+  public removeUrlFilter(attributes: FilterAttribute[], filter: Filter): void {
+    const filters = this.getUrlFilters(attributes);
+
+    const foundIndex = filters.findIndex(f => filter.field === f.field);
```
aaron-steinfeld, line 51: Why findIndex + splice instead of a plain filter? As far as I can tell, this whole method just collapses down to:
`this.setUrlFilters(this.getUrlFilters(attributes).filter(f => filter.field === f.field))`

projects/components/src/filter-button/filter-button.service.ts
```diff
@@ -37,17 +45,29 @@ export class FilterButtonService {
...
+  }
+
   private setUrlFilters(filters: Filter[]): void {
     this.navigationService.addQueryParametersToUrl({
       [FilterButtonService.FILTER_QUERY_PARAM]: filters.length === 0 ? undefined : filters.map(f => f.urlString)
     });
   }
 
-  private getUrlFilters(attributes: FilterAttribute[]): Filter[] {
+  public getUrlFilters<T>(attributes: FilterAttribute[]): Filter<T>[] {
```
aaron-steinfeld, line 66: If we want to move the filters out of the URL (which I think we should), then it'd be nice to abstract the add/remove/read methods to remove references to URLs.

projects/components/src/table/table.component.ts
```diff
@@ -73,7 +73,7 @@ import {
               [columnConfig]="columnDef"
               [index]="index"
               [sort]="columnDef.sort"
-              [values]="this.getAvailableValues(columnDef) | async"
+              [values]="this.getAvailableValues(columnDef)"
```
aaron-steinfeld, line 76: This still has the same issue as before - it's returning a new reference on each invocation

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -59,19 +62,45 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
+    rows.forEach(row => {
+      Object.entries(row).forEach((keyValueTuple: [string, unknown]) => {
+        const key = keyValueTuple[0];
+        const value = keyValueTuple[1];
+
+        if (!this.cachedValues.has(key)) {
+          this.cachedValues.set(key, new Set());
+        }
+
+        const set = this.cachedValues.get(key);
```
anandtiwary, line 97: This would always be defined right due to line 94? If yes, we don't need the else part in line 99.

projects/components/src/filter-button/in-filter-button.component.ts
arjunlalb: Can you attach a screenshot / recording? I'm getting the impression that the only difference with the multi-select is that a label isn't used. If that is the case, we can make it optional and reduce the redundancy.

arjunlalb: Can you attach a screenshot / recording? I'm getting the impression that the only difference with the multi-select is that a label isn't used. If that is the case, we can make it optional and reduce the redundancy.

projects/components/src/table/data/table-cdk-data-source.ts
jake-bassett: True, and good catch.

The reason I have the check is because the typing allows for `set` to be undefined. I could add an assertion, but saw no harm in leaving the check.

anandtiwary: okay

projects/components/src/filter-button/in-filter-button.component.ts
anandtiwary: I think this one is unresolved yet. Jake can you please check?

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -59,23 +62,45 @@ export class TableCdkDataSource implements DataSource<TableRow> {
...
   public disconnect(): void {
     this.rowsChange$.complete();
-    this.loadingState.complete();
+    this.loadingStateSubject.complete();
   }
 
   public getValues(columnConfig: TableColumnConfig): unknown[] {
-    return [...new Set(this.cachedRows.map(row => row[columnConfig.field]))].sort(sortUnknown);
+    return this.cachedValues.has(columnConfig.field)
+      ? [...this.cachedValues.get(columnConfig.field)!].sort(sortUnknown)
```
aaron-steinfeld, line 77: Why not sort on set rather than every get?

projects/components/src/table/data/table-cdk-data-source.ts
aaron-steinfeld: Seems weird to me. If we want to follow the typing, no reason for 93-95, else we should assert based on our check.

projects/components/src/table/data/table-cdk-data-source.ts
jake-bassett: Smart man. Good catch!

projects/components/src/table/data/table-cdk-data-source.ts
jake-bassett: Eh, that's fine. Need to fix the other comment now so I can clean this up.

projects/components/src/table/data/table-cdk-data-source.ts
jake-bassett: Ah, I actually remember now. This is a set so its unsorted.

What I can do instead though is use a set to build the unique list but store an array in the map instead.

aaron-steinfeld: Easy confusion, but JS Sets are actually sorted according to the spec, so up to you.
> Set objects are collections of values. You can iterate through the elements of a set in insertion order. A value in the Set may only occur once; it is unique in the Set's collection.

jake-bassett: Correct, but they can't be sorted in place. I'd still have to sort prior to adding all of the values to the set. So same solution of building an array and sorting. It just changes when.

I'll build the set first to collect only unique, then transfer to `cachedValues` as a sorted array.

projects/components/src/filter-button/filter-button.service.ts
jake-bassett: Agreed. Will do that as part of that effort.

projects/components/src/filter-button/filter-button.service.ts
jake-bassett: Fixed.

projects/components/src/filter-button/in-filter-button.component.ts
jake-bassett: Thanks. This was buried under a "Viewed" file so I missed it. Sorry about that.

There are a few incompatibilities with the current multiselect. Yes, the label is one but also the long term vision for this dropdown is more than just checkboxes. We may add additional column actions like removing it to the same dropdown.

I've attached the mock for the current usage, but this will grow.

Screen Shot 2020-08-26 at 1 54 42 PM

arjunlalb: gotcha.

PR #110, opened 2020-08-26
Responsive components

anandtiwary:


projects/components/src/not-found/not-found.component.ts
```diff
@@ -8,9 +8,7 @@ import { ButtonStyle } from '../button/button';
   styleUrls: ['./not-found.component.scss'],
   template: `
     <div class="not-found-container fill-container">
-      <div class="not-found-message">
-        The requested page is not available
-      </div>
+      <div class="not-found-message">The requested page is not available</div>
```
anandtiwary, line 11: Prettier picked this up in my change when i ran it for the entire repo.

projects/components/src/overlay/sheet/sheet.ts
```diff
@@ -8,5 +8,6 @@ export const enum SheetSize {
   Small = 'small',
   Medium = 'medium',
   Large = 'large',
-  ExtraLarge = 'extra-large'
+  ExtraLarge = 'extra-large',
+  Responsive = 'responsive'
```
aaron-steinfeld, line 12: While responsive, this seems like it should be called halfwidth or something. Responsive itself gives no indication of size like the others do.

anandtiwary: Yeah. I knew you would ask this. I see responsive as a separate property than sheet size. You can have small size and have it be responsive. I discussed this with Michael as well, and there is just one use case at the moment so deferred this change for later. Like right now, we are only using one size which is Large everywhere.

anandtiwary: Infact talking to him today, we decided to keep it at 60%. 

how about we expose a new property as mentioned above? And how about we do it separately at a later stage ? If you agree, would you still want to rename 'Responsive' to idk `ResponsiveAt60Percent` ?

aaron-steinfeld: Either something like ResponsiveMedium (mirroring the others) or making responsive a separate config like you said.

anandtiwary: We would need to do something like based on some breakpoints. Below one is for Extra large Sheet. So this needs more inputs from Michael and I would take it up probably next week, if its okay.
```
window size: 480, sheet size: 320… = 66%
window size: 960, sheet size: 600… = 60%
window size: 1440, sheet size: 840 = ~60%
window size: 1920, sheet size: 1280 = ~66%
```

anandtiwary: And now mostly responsive will be in-built and will be available to all sizes ootb.

anandtiwary: Renamed this to `ResponsiveExtraLarge` to match the size.

PR #111, opened 2020-08-27
feat: adding a darker modal backdrop

anandtiwary:


projects/components/src/popover/popover.service.ts
```diff
@@ -25,7 +25,7 @@ export class PopoverService {
     const overlayRef = this.overlay.create({
       positionStrategy: initialPositionStrategy,
       hasBackdrop: options.backdrop,
-      backdropClass: 'cdk-overlay-transparent-backdrop'
+      backdropClass: options.backdrop ? 'modal-overlay-backdrop' : 'cdk-overlay-transparent-backdrop'
```
aaron-steinfeld, line 28: This turns all backdrops dark. If I recall, backdrop class is unused if hasBackdrop is false, so the transparent class would never be used here. We would need to migrate backdrop the config boolean to an enum to the effect of [TRANSPARENT, OPAQUE, NONE] (or use undefined to indicate none)

anandtiwary: I didn't see any visual difference between None and transparent. Is there one?

aaron-steinfeld: Visually no, it's a behavior difference - with a transparent backdrop, interaction is blocked with anything outside the popover (and generally we'd close on a click on the backdrop). At least, that was the initial implementation as I recall.

anandtiwary: oh I see. Makes sense. Will add the new enum state

projects/components/src/popover/popover.service.ts
```diff
@@ -78,4 +78,20 @@ export class PopoverService {
...
   }
+
+  private hasBackdrop(backdrop?: PopoverBackdrop): boolean {
+    return backdrop === PopoverBackdrop.Opaque || backdrop === PopoverBackdrop.Transparent;
+  }
+
+  private getBackdropClass(backdrop?: PopoverBackdrop): string {
+    switch (backdrop) {
+      case PopoverBackdrop.Transparent:
+        return 'cdk-overlay-transparent-backdrop';
```
aaron-steinfeld, line 89: not sure what's in this, but does it make sense to migrate to our own class for consistency? how much can a transparent backdrop have in it?

anandtiwary: I don't think there is any benefit. It sets the opacity. I think we can leave it as it is till we modify the style.

PR #113, opened 2020-08-28
fix: cell click with handler stops prop to row

jake-bassett:


projects/components/src/table/cells/data/table-data-cell-renderer.component.ts
```diff
@@ -107,8 +107,13 @@ export class TableDataCellRendererComponent implements OnInit {
...
   }
 
-  public onClick(): void {
-    this.columnConfig && this.columnConfig.onClick && this.columnConfig.onClick(this.rowData!, this.columnConfig);
+  public onClick(event: MouseEvent): void {
+    const hasClickHandler = this.columnConfig && this.columnConfig.onClick;
+
+    if (hasClickHandler) {
+      this.columnConfig!.onClick!(this.rowData!, this.columnConfig!);
+      event.stopPropagation();
```
anandtiwary, line 115: Wouldn't this block click propagation for row selection?

aaron-steinfeld: I'd think so - what's the bug we're seeing here?

jake-bassett: It blocks it only if you click specifically on the div on line 27 AND it has a click handler. This way the most specific handler gets the click event and stops it so there aren't two actions (one for the cell and one for the row).

The bug is right now if you click in a cell and it has a handler the row also gets the click. You can see this in multiple places, but its most obvious with waterfall where if you try to expand/collapse the row selection also opens the side panel.

aaron-steinfeld: Got it - is that clickbox limited to the rendered content, or it full cell width?

anandtiwary: okay. This makes sense to me, however, i think we should reduce the cell area to as minimum as possible. This is because, on security screens, we add both the click and row handler. And I have noticed the cell area sometimes take a lot of white spaces especially width.

anandtiwary: That makes it difficult to click on the row as now you only have to click on the whitespace above/below the cell text.

aaron-steinfeld: I'd expect the cell clickbox to be full height but match the rendered content width.

anandtiwary: Full height would definitely not work as some cells are two small in its width. So row click-ability would become more difficult. I know Michael's opinion on this as well when i discussed with him last, the cell click area would be limited to the rendered cell text. All the other white space would be used for row click-ability.  Same behavior we are supporting on the cards.

jake-bassett: Screen Shot 2020-08-28 at 1 05 07 PM

Here is the size of the clickable element. I think this is the best compromise on space usage.

anandtiwary: Yeah. Per UX mocks, the empty space horizontally should invoke row handler

PR #114, opened 2020-08-28
feat: add units to Duration in explorer

jake-bassett:


projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
```diff
@@ -63,10 +66,12 @@ export class TracesGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
...
         const data = spec.extractFromServerData(rawResult);
 
-        return of({
-          alias: alias,
-          data: data
-        });
+        return this.resultUnits(spec).pipe(
+          map(units => ({
+            alias: alias,
+            data: units !== undefined ? { units: units, value: data } : data
```
aaron-steinfeld, line 72: I'm really not a fan of this since it's changing the shape of the response dynamically based on something unknowable on the request side (that is, the metadata associated with a key), it makes the code super brittle. If unit is not defined, it returns a primitive and if it is defined an object? Can we precede this with a change to make attributes values, like metric values currently, represented by an object containing a value field and optionally units?

jake-bassett: Agreed; not a fan either. I like your suggested solution but that is a much bigger change. We already do this same (not great) pattern for SpansGraphQlQueryHandlerService and others. Can we come back and refactor later? 

I want to fix all of our unit enrichment code. Lots of it is repeated. In reality, I want to find a much better way of marrying data with metadata in general.

aaron-steinfeld: Yep, if this is consistent with spans lets go for it.

projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
```diff
@@ -81,6 +86,12 @@ export class TracesGraphQlQueryHandlerService implements GraphQlQueryHandler<Gra
       })
     );
   }
+
+  private resultUnits(specification: Specification): Observable<string | undefined> {
+    return this.metadataService
+      .getAttribute('API_TRACE', specification.name)
```
aaron-steinfeld, line 92: the scope should come off the requested trace type

projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
jake-bassett: Thanks

projects/distributed-tracing/src/shared/graphql/request/handlers/traces/traces-graphql-query-handler.service.ts
jake-bassett: Fixed.

PR #120, opened 2020-08-31
fix(table): sortable, titleTooltip, clear selections on page change

skjindal93:


projects/components/src/table/table.component.ts
```diff
@@ -603,6 +616,10 @@ export class TableComponent
         }
       : undefined;
   }
+
+  private isColumnSortable(columnConfig: TableColumnConfig): boolean {
+    return columnConfig.sortable === undefined || columnConfig.sortable;
```
arjunlalb, line 621: shouldn't this be `value !== undefined && sortable` ? also, is `sortable` a valid property? shouldn't it be `sort` ?

skjindal93: Actually, I wanted it to have sort by default, even if someone does not add sortable property in the column config

projects/components/src/table/table.component.ts
```diff
@@ -345,10 +352,12 @@ export class TableComponent
   }
 
   public onHeaderCellClick(columnConfig: TableColumnConfig): void {
-    this.updateSort({
-      column: columnConfig,
-      direction: this.getNextSortDirection(columnConfig.sort)
-    });
+    if (this.isColumnSortable(columnConfig)) {
```
aaron-steinfeld, line 355: Is sortable not used anywhere any more? must be a regression. We should also change the styling for not sortable so it doesn't have a hover pointer.

projects/components/src/table/table.component.ts
```diff
@@ -562,6 +571,10 @@ export class TableComponent
         [TableComponent.PAGE_SIZE_URL_PARAM]: pageEvent.pageSize
       });
     }
+
+    this.selections = [];
+    this.selectionsChange.emit(this.selections);
```
aaron-steinfeld, line 576: This should only emit if there were previously selections - that is, it changed.

skjindal93: Just for optimization, right?

aaron-steinfeld: For correctness - this is a change emitter, it should only emit on change.

projects/components/src/table/table.component.ts
skjindal93: Right. The sortable was behaving as a no-op (_as in it was always sorting, irrespective of the value_)

projects/components/src/table/data/table-cdk-column-util.ts
```diff
@@ -17,5 +17,11 @@ export namespace TableCdkColumnUtil {
   export const unsortOtherColumns = (sortedColumn: TableColumnConfig, otherColumns: TableColumnConfig[]): void =>
     otherColumns
       .filter(column => column.field !== sortedColumn.field)
-      .forEach(filteredColumn => (filteredColumn.sort = undefined));
+      .forEach(filteredColumn => {
+        filteredColumn.sort = undefined;
+        filteredColumn.sortable = false;
```
anandtiwary, line 22: Should we set `sortable = false`? Wouldn't this take out the sort feature of an existing column?

skjindal93: Shit. Yes. Nice catch!

PR #122, opened 2020-09-01
refactor: pull parsing code out of renderers

jake-bassett:


projects/components/src/table/cells/data-parsers/table-cell-timestamp-parser.ts
```diff
@@ -0,0 +1,18 @@
...
+@TableCellParser({
+  type: CoreTableCellParserType.Timestamp
+})
+export class TableCellTimestampParser extends TableCellParserBase<CellData, CellData, CellData> {
+  public parseValue(cellData: CellData): CellData {
+    return cellData;
+  }
+
+  public parseFilterValue(cellData: CellData): CellData {
+    return this.parseValue(cellData);
```
aaron-steinfeld, line 14: should this be in the base class?

projects/components/src/table/cells/table-cell-parser-base.ts
```diff
@@ -0,0 +1,17 @@
+import { TableRow } from '../table-api';
+import { CoreTableCellParserType } from './types/core-table-cell-parser-type';
+
+export abstract class TableCellParserBase<TCellData, TValue, TFilterValue> {
+  public static readonly type: CoreTableCellParserType;
+
+  public abstract parseValue(cellData: TCellData, rowData: TableRow): TValue;
+  public abstract parseFilterValue(cellData: TCellData): TFilterValue;
```
aaron-steinfeld, line 8: shouldn't this default to just parseValue? Seems like that's its implementation almost everywhere.

projects/components/src/table/cells/table-cell-parser-lookup.service.ts
```diff
@@ -0,0 +1,27 @@
+import { Injectable } from '@angular/core';
+import { TableCellParserConstructor } from './table-cell-parser';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TableCellParserLookupService {
+  private readonly parsers: Map<string, TableCellParserConstructor<unknown, unknown, unknown>> = new Map();
+
+  public register(parser: TableCellParserConstructor<unknown, unknown, unknown>): void {
```
aaron-steinfeld, line 10: Can collapse these two apis with a rest arg

projects/components/src/table/cells/table-cell-parser.ts
```diff
@@ -0,0 +1,26 @@
...
+  extends Type<TableCellParserBase<TCellData, TValue, TFilterValue>>,
+    TableCellParserMetadata {}
+
+/*
+ * @TableCellParser decorator is used to configure type for cell parsers
+ */
+// tslint:disable-next-line:only-arrow-functions
+export function TableCellParser(tableCellParserMetadata: TableCellParserMetadata): TableCellParserDecorator {
+  return (constructor: TableCellParserConstructor<unknown, unknown, unknown>): void => {
+    constructor.prototype.type = constructor.type = tableCellParserMetadata.type;
```
aaron-steinfeld, line 18: please no double assignment. Also, where do we use this on the prototype?

projects/components/src/table/cells/table-cell-renderer-lookup.service.ts
```diff
@@ -1,83 +1,60 @@
...
-    if (type === undefined) {
-      return this.defaultOrError();
-    }
-
+  public lookup(type: string): TableCellRendererConstructor {
     if (!this.renderers.has(type)) {
-      this.loggerService.warn(`Table cell renderer by the name '${type}' not found. Using default.`);
-
-      return this.defaultOrError();
+      throw Error(`Table cell parser of type '${type}' not registered.`);
```
aaron-steinfeld, line 29: No default any more?

projects/distributed-tracing/src/shared/components/table/data-cell/metric/metric-table-cell-parser.ts
```diff
@@ -0,0 +1,44 @@
+import { TableCellParser, TableCellParserBase } from '@hypertrace/components';
+import { MetricAggregation } from '../../../../graphql/model/metrics/metric-aggregation';
+import { TracingTableCellType } from '../../tracing-table-cell-type';
+
+@TableCellParser({
+  type: TracingTableCellType.Metric
+})
+export class MetricTableCellParser extends TableCellParserBase<number, number, number> {
+  public parseValue(cellData: CellData): number {
+    return Math.round(this.extractValue(cellData)!);
```
aaron-steinfeld, line 10: This is pre-existing, but why are we rounding? We show decimals in places, right?

projects/components/src/table/cells/table-cell-renderer.ts
```diff
@@ -1,22 +1,32 @@
...
-
+/*
+ * @TableCellRenderer decorator is used to configure type, alignment, parser for cell renderers
+ */
 // tslint:disable-next-line:only-arrow-functions
 export function TableCellRenderer(tableCellRendererMetadata: TableCellRendererMetadata): TableCellRendererDecorator {
   return (constructor: TableCellRendererConstructor): void => {
-    constructor.type = tableCellRendererMetadata.type;
-    constructor.alignment = tableCellRendererMetadata.alignment;
+    constructor.prototype.type = constructor.type = tableCellRendererMetadata.type;
```
aaron-steinfeld, line 20: Why on the prototype? Feels hacky. Does the type reflect this?

projects/components/src/table/table-api.ts
```diff
@@ -1,14 +1,18 @@
...
 import { Observable } from 'rxjs';
 import { FilterAttribute } from '../filter-bar/filter-attribute';
+import { TableCellParserConstructor } from './cells/table-cell-parser';
+import { TableCellRendererConstructor } from './cells/table-cell-renderer';
 import { TableCellAlignmentType } from './cells/types/table-cell-alignment-type';
 
 export interface TableColumnConfig {
   field: string;
+  display?: string;
+  renderer?: TableCellRendererConstructor;
```
aaron-steinfeld, line 11: In other places we seem to assume renderer and parser are defined. This feels like we're mixing up internal and external representations. We shouldn't require these in addition to display, since display encompasses both of them.

projects/components/src/table/cells/data-renderers/table-data-cell-renderer.component.ts
```diff
@@ -117,6 +111,6 @@ export class TableDataCellRendererComponent implements OnInit {
   }
 
   public parseValue(): unknown {
-    return this.componentRef?.instance.parseValue(this.cellData, this.rowData);
+    return new this.columnConfig!.parser!().parseFilterValue(this.cellData);
```
aaron-steinfeld, line 114: I'll keep saying it - every time you have to use `!`, it's indicative of a problem.

projects/components/src/table/cells/table-cell-renderer-base.ts
```diff
@@ -0,0 +1,56 @@
...
+    @Inject(TABLE_CELL_DATA) private readonly cellData: TCellData,
+    @Inject(TABLE_ROW_DATA) private readonly rowData: TableRow,
+    @Inject(TABLE_COLUMN_INDEX) private readonly index: number
+  ) {
+    this.clickable = this.columnConfig.onClick !== undefined;
+    this.isFirstColumn = this.index === 0;
+  }
+
+  public ngOnInit(): void {
+    const cellParser = new this.columnConfig.parser!() as TableCellParserBase<TCellData, TValue, unknown>;
```
aaron-steinfeld, line 38: Because we manually `new` this, we're limiting the parsers to be non-injectable and take no constructor args, which is non-obvious and not reflected (or possible to reflect, because it breaks a polymorphic contract - the base type shouldn't enforce anything about the child types constructor) in the types.

projects/components/src/table/data/table-cdk-data-source.ts
```diff
@@ -89,10 +90,14 @@ export class TableCdkDataSource implements DataSource<TableRow> {
         const key = keyValueTuple[0];
         const value = keyValueTuple[1];
 
+        const columnConfig = this.columnConfigs.get(key)!;
+        const parser = new columnConfig.parser!();
```
aaron-steinfeld, line 94: Rather than new'ing this everywhere, can we just put an instance of the parser on the column config?

projects/distributed-tracing/src/shared/dashboard/widgets/table/table-widget-column.model.ts
```diff
@@ -91,13 +95,24 @@ export class TableWidgetColumnModel {
...
   public sort?: TableSortDirection;
 
+  @ModelInject(TableCellRendererLookupService)
+  private readonly tableCellRendererLookupService!: TableCellRendererLookupService;
+
+  @ModelInject(TableCellParserLookupService)
+  private readonly tableCellParserLookupService!: TableCellParserLookupService;
+
   public asTableColumnDef(filterAttribute?: FilterAttribute): SpecificationBackedTableColumnDef {
+    const rendererConstructor = this.tableCellRendererLookupService.lookup(this.display);
```
aaron-steinfeld, line 105: Why not leave this responsibility to the table? Are we expecting table columns that are not defined through widgets to always provide a parser and renderer?

projects/components/src/table/cells/table-cell-parser-base.ts
jake-bassett: Currently there are 9 cell renderers that don't require override and 6 that do. Since the majority of the renderers that _don't_ require the override and can just use `parseValue` directly are the primitive types I think I'd like to keep it abstract. Most, if not all, renderers we add from now on will be composite and require the `parseFilterValue` method.

I think having users see they need to implement it will prevent them from forgetting it and causing a filtering issue.

projects/components/src/table/cells/table-cell-renderer-lookup.service.ts
jake-bassett: No. Default is set in the model now if nobody overrides it. Simplifies things a bit.

projects/distributed-tracing/src/shared/components/table/data-cell/metric/metric-table-cell-parser.ts
jake-bassett: I think this was right for Duration, but maybe not for others.

projects/components/src/table/cells/table-cell-renderer-lookup.service.ts
aaron-steinfeld: Does it mean using the table directly anywhere outside a dashboard is required to be explicit?

jake-bassett: Hmm. Good point. We probably don't want to require that.

Going to change this back to have a default.

projects/components/src/table/table-api.ts
```diff
@@ -1,26 +1,32 @@
...
   sort?: TableSortDirection;
   visible?: boolean;
   sortable?: boolean;
+  filterable?: boolean;
   alignment?: TableCellAlignmentType;
   width?: number | string;
   onClick?(row: TableRow, column: TableColumnConfig): void;
 }
 
+export interface TableColumnConfigExtended extends TableColumnConfig {
```
aaron-steinfeld, line 23: What happens if we don't export this? The idea of having an internal version of the column config is to avoid expoing implementation details, but this interface just exploded all over this commit and makes it even more confusing. At the least, anything that gets plugged into the table - i.e. parsers, renderers, columns etc. - should be working with a regular column, and anything that gets emitted should be a regular column.

projects/components/src/table/cells/data-renderers/table-data-cell-renderer.component.ts
```diff
@@ -85,32 +84,37 @@ export class TableDataCellRendererComponent implements OnInit {
...
   public onClick(event: MouseEvent): void {
-    const hasClickHandler = this.columnConfig && this.columnConfig.onClick;
+    if (this.columnConfig === undefined || this.rowData === undefined) {
+      throw Error(`Undefined columnConfig or rowData`);
+    }
 
-    if (hasClickHandler) {
-      this.columnConfig!.onClick!(this.rowData!, this.columnConfig!);
+    if (this.isClickHandlerDefined(this.columnConfig.onClick)) {
+      this.columnConfig.onClick(this.rowData, this.columnConfig);
```
aaron-steinfeld, line 103: Can we just use the safe access rather than the runtime type check? Is there a scenario where this could be anything besides undefined or a function? `this.columnConfig?.onClick(this.rowData, this.columnConfig)`

projects/components/src/table/cells/data-renderers/table-data-cell-renderer.component.ts
```diff
@@ -19,22 +19,22 @@ import { TableCellAlignmentType } from '../types/table-cell-alignment-type';
...
       </ng-container>
 
       <ng-template #filterButton>
         <htc-filter-button
           class="filter-button"
           [metadata]="this.metadata"
-          [attribute]="this.columnConfig?.filterAttribute"
-          [value]="this.parseValue()"
+          [attribute]="this.columnConfig?.attribute"
+          [value]="this.parseFilterValue()"
```
aaron-steinfeld, line 37: this will be reinvoked every cycle

projects/components/src/table/table.service.ts
```diff
@@ -0,0 +1,59 @@
...
+})
+export class TableService {
+  private static readonly STATE_ATTRIBUTE: FilterAttribute = {
+    name: '$$state',
+    displayName: 'Row State',
+    type: '$$state' as FilterType
+  };
+
+  public constructor(
+    private readonly injector: Injector,
```
aaron-steinfeld, line 21: Since this is a root service, it will get a root injector rather than the injector in the context of the current table.

projects/components/src/table/table.service.ts
```diff
@@ -0,0 +1,59 @@
...
+        columnConfig.display !== undefined ? columnConfig.display : CoreTableCellRendererType.Text
+      );
+
+      const parserConstructor = this.tableCellParserLookupService.lookup(rendererConstructor.parser);
+
+      return {
+        ...columnConfig,
+        attribute: attribute,
+        renderer: rendererConstructor,
+        parser: new parserConstructor(this.injector),
```
aaron-steinfeld, line 46: It feels cleaner to make the parsers just services and inject instances of them, rather than plain classes that we're tossing instances of an injector. They're stateless right?

projects/components/src/table/table.component.ts
```diff
@@ -419,20 +416,20 @@ export class TableComponent
     return this.hasExpandableRows() ? index - 1 : index;
   }
 
-  private buildColumnConfigs(): TableColumnConfig[] {
-    if (!this.columnConfigs) {
+  private buildColumnConfigExtendeds(): TableColumnConfigExtended[] {
```
aaron-steinfeld, line 419: `buildExtendedColumnConfigs`

projects/components/src/table/cells/data-parsers/table-cell-timestamp-parser.ts
jake-bassett: I see the the value in having a default implementation, but as mentioned in the other comment, would prefer to keep the method abstract.

projects/components/src/table/cells/table-cell-parser-lookup.service.ts
jake-bassett: Done.

projects/components/src/table/cells/table-cell-parser.ts
jake-bassett: Fixed.

projects/components/src/table/cells/table-cell-renderer-lookup.service.ts
jake-bassett: Fixed.

projects/components/src/table/cells/table-cell-renderer.ts
jake-bassett: Fixed.

projects/components/src/table/cells/table-cell-injection.ts
```diff
@@ -0,0 +1,36 @@
+import { InjectionToken, Injector } from '@angular/core';
+import { TableColumnConfig, TableColumnConfigExtended, TableRow } from '../table-api';
+
+export const TABLE_COLUMN_CONFIG: InjectionToken<TableColumnConfig> = new InjectionToken('TABLE_COLUMN_CONFIG');
+export const TABLE_COLUMN_INDEX: InjectionToken<TableColumnConfig> = new InjectionToken('TABLE_COLUMN_INDEX');
+export const TABLE_CELL_DATA: InjectionToken<unknown> = new InjectionToken('TABLE_CELL_DATA');
+export const TABLE_ROW_DATA: InjectionToken<unknown> = new InjectionToken('TABLE_ROW_DATA');
+
+export const createInjector = (
```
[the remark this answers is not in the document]
jake-bassett: Fixed.

projects/components/src/table/cells/table-cell-parser-base.ts
```diff
@@ -1,9 +1,12 @@
+import { Injector } from '@angular/core';
 import { TableRow } from '../table-api';
 import { CoreTableCellParserType } from './types/core-table-cell-parser-type';
 
 export abstract class TableCellParserBase<TCellData, TValue, TFilterValue> {
   public static readonly type: CoreTableCellParserType;
 
+  public constructor(protected readonly injector: Injector) {}
```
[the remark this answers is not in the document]
jake-bassett: Yeah, this is annoying.

The only place we need this right now is in `BackendIconTableCellParser` because it needs to inject `EntityIconLookupService`.

projects/components/src/table/cells/table-cell-renderer-base.ts
```diff
@@ -25,17 +20,17 @@ export abstract class TableCellRendererBase<TCellData, TValue = TCellData> imple
...
-    @Inject(TABLE_COLUMN_INDEX) private readonly index: number
+    @Inject(TABLE_ROW_DATA) private readonly rowData: TableRow
   ) {
     this.clickable = this.columnConfig.onClick !== undefined;
     this.isFirstColumn = this.index === 0;
   }
 
   public ngOnInit(): void {
-    const cellParser = new this.columnConfig.parser!() as TableCellParserBase<TCellData, TValue, unknown>;
+    const cellParser = this.columnConfig.parser as TableCellParserBase<TCellData, TValue, unknown>;
```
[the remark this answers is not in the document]
jake-bassett: Fixed.

projects/components/src/table/table-api.ts
```diff
@@ -1,26 +1,32 @@
...
 import { FilterAttribute } from '../filter-bar/filter-attribute';
-import { TableCellParserConstructor } from './cells/table-cell-parser';
+import { TableCellParserBase } from './cells/table-cell-parser-base';
 import { TableCellRendererConstructor } from './cells/table-cell-renderer';
 import { TableCellAlignmentType } from './cells/types/table-cell-alignment-type';
 
 export interface TableColumnConfig {
-  field: string;
+  field: string; // This is the unique name for the column (usually same as name except for composite fields)
+  name?: string; // Attribute name (for composite fields this should be the filterable attribute)
```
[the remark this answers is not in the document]
jake-bassett: Fixed... kind of.

Cleared up the naming a bit. I do think there is a place for both `id` and `name`. We need `name` not only for filtering but also for sorting.

projects/components/src/table/table-api.ts
jake-bassett: Fixed.

projects/components/src/table/cells/data-renderers/table-data-cell-renderer.component.ts
jake-bassett: I simplified it so we don't need the ts ignore comment.

projects/components/src/table/cells/data-renderers/table-data-cell-renderer.component.ts
jake-bassett: Fixed.

projects/components/src/table/table.service.ts
jake-bassett: Correct. Root injector is fine. I renamed to clear up any assumptions.

projects/components/src/table/table.service.ts
jake-bassett: Eh, I'm not so sure I agree on this one. But when I move parsers out of table I will explore the idea of making them injectable at that time.

projects/components/src/table/table.component.ts
jake-bassett: This is the hill I die on!

I actually had it as `buildExtendedColumnConfigs` originally, but changed it to match the name of the type. It hurts me from an English language standpoint, but now it matches our usual naming conventions! ¯\_(ツ)_/¯

projects/components/src/table/cells/table-cell-parser-base.ts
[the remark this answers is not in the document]
aaron-steinfeld: Oh this is because we treat it as the base class and construct it. Will finish this thought in that thread.

projects/components/src/table/table.service.ts
aaron-steinfeld: In terms of making them injectable, wouldn't it just be marking them as `@Injectable` ? Then, we don't need to new it, don't need the base constructor param since each parser is independent, etc.
Then here, instead of newing, `injector.get(parserConstructor)`. But I guess that's predicated on thinking that's the right path.

projects/components/src/table/table.component.ts
aaron-steinfeld: Ooh good hill - I'd argue that shows the name should be... `ExtendedTableColumnConfig` , which is also more englishy.

projects/components/src/table/table.service.ts
aaron-steinfeld: Feels like it's making an assumption on the parsers, although I agree that for all practical applications it should be the same.

aaron-steinfeld (approved):

projects/components/src/table/table.component.ts
jake-bassett: Hah. I had that first as well, but then didn't like them not being alphabetically adjacent in imports. I hate English sometimes.

projects/components/src/table/table.service.ts
jake-bassett: I'll definitely explore. Not happy with the `new` either, but it seems the most light weight. Maybe trying to switch things over will convince me its not too bad.

aaron-steinfeld: `new ` and `.get` is the same number of chars. just sayin'

PR #126, opened 2020-09-03
refactor: making sheet and modal coexist

anandtiwary:


projects/components/src/overlay/overlay.service.ts
```diff
@@ -13,16 +13,22 @@ import { SheetOverlayComponent } from './sheet/sheet-overlay.component';
   providedIn: 'root'
 })
 export class OverlayService implements OnDestroy {
-  private activePopover?: PopoverRef;
-  private closeSubscription?: Subscription;
+  private activeSheetPopover?: PopoverRef;
```
aaron-steinfeld, line 16: Should these just be separate services at this point? How much is shared?

anandtiwary: We can make it separate as well. It just delegates to popover. But it doesn't seem wrong for them to be in the same service either.

anandtiwary: Will prefer to keep it in the same service for now.

jake-bassett (approved):

PR #127, opened 2020-09-03
fix: table console error

jake-bassett:


projects/components/src/table/table.component.ts
```diff
@@ -425,7 +425,7 @@ export class TableComponent
       colConfigs = [this.multiSelectRowColumnConfig, ...this.columnConfigs];
     }
 
-    return this.tableService.buildExtendedColumnConfigs(colConfigs, this.dataSource, this.metadata);
+    return this.tableService.buildExtendedColumnConfigs(colConfigs, this.dataSource, this.metadata || []);
```
aaron-steinfeld, line 428: `this.metadata` looks like it can be undefined according to its type and it wasn't a compile error, so is there an issue in the function that doesn't work with undefined?

jake-bassett: What's actually happening is the first round of change detection is setting `metadata` to `null` and that is getting through. I'm guessing a consumer of the table is setting the metadata input but the value isn't populated yet, so we pick up the null.

Open to how best to fix this.

aaron-steinfeld: ah yeah, it's kind of an unspoken agreement that inputs could be either null or undefined, so if we're only checking for undefined, that'd do it. Our own async constructs stick to undefined, but angular's ` | async` switches between the two.

aaron-steinfeld (approved):

